PR #170, opened 2020-06-30
Json support

matteofigus:
*Description of changes:*
PR for adding JSON support with minimal featureset to the main branch.

### Architecture summary

The architecture is left pretty much unchanged. The Find phase works as expected when mapping from Glue Tables. 

For the Forget Phase I tested 2 main approaches: first using Apache Arrow for parsing and recomposing the files, then native JSON libraries. The approach I came up is mixed:
1) I use native python libraries for gzip and json to read lines, identify rows and copy over to a new buffer
2) I use Arrow's `BufferOutputStream` and `CompressedOutputStream` to build up the new object

The reason I went native for the parsing, is that Arrow is quite opinionated on the way serialisation is done. I found a bug where a string recognised as date (ex. `2020-01-01`) would be automatically converted to Arrow's date, and then stringified back as `2020-01-01T00:00:00` rather than original format (https://arrow.apache.org/docs/python/json.html#automatic-type-inference).
The second reason is that I found Arrow's strong typing would require each row to adhere to the same schema. For instance, two rows like `{"a": 123, "b":{"c": 123}}` and `{"a":234, "b":{"d":444}}` would cause parsing issues. The only way to cleanly solve both issues seems to be passing the schema rather than allowing Arrow/Pandas to infer that from the content, but the issue is that I would need to create a mapper from Glue syntax to Arrow, and that is really really complex. I did a spike and it involved lots of code for simple objects, and turned to be very challenging for a real dataset.

With my approach, instead, I parse row by row, which seems to be very memory/cpu efficient and allows arbitrary schemas. I found that using Arrow's stream utilities to build up the new file was quite convenient, as it allowed writing the ouput correctly and limiting changes to the `main.py` code. The `CompressedOutputStream` supports many compression formats, which could be useful if in the future I need to add more other than `gzip`.

### SerDes and parameters

I tested with complex datamappers, for instance using the [free Github Dataset](https://s3.console.aws.amazon.com/s3/buckets/awsglue-datasets/examples/scala-blog/githubarchive/data/?region=us-east-1) and using multiple column identifiers such as `actor.login, payload.pull_request.merged_by.login, payload.issue.user.login, payload.issue.assignee.login, payload.issue.milestone.creator.login, payload.comment.user.login, payload.forkee.owner.login, payload.release.author.login, payload.member.login, org.login`, and all worked fine. I made a 100GB based on that and reported the numbers as new cost scenario.

I am adding support for both SerDe supported by glue: the hive one and OpenX one: https://docs.aws.amazon.com/athena/latest/ug/json-serde.html

As you can see from the OpenX, there are a number of SerDe params that somehow affect the column mapping and the Forget Phase. I decided to start small and stay simple for this PR, and I'll add more later on:
* `ignore.malformed.json`: It allows only FALSE or empty (default is FALSE). I actually wonder if we would ever want to allow this to be true given this solution's goal.
* `dots.in.keys`: Athena by default doesn't allow dots in keys. If you have dots in your keys, you need to setup this to `TRUE` and then athena converts dots to underscores. We could potentially support this passing down the information to Fargate, but at the moment I just added a check to prevent this to be TRUE (default is FALSE).
* `case.insensitive` and `mapping`: Athena/Glue automatically converts any key to lowercase, `userId` in file, for instance, becomes `userid` in table. With `case.insensitive`=FALSE you can setup your own mappings, allowing you to handle complex scenarios such as having both a `URL` and `Url` keys which wouldn't allow Athena to proper serialise. I decided, for this first iteration, to disallow mapping of any kind, and in the Forget, I added a little piece of code to navigate the object's tree by lowercasing the found keys (but obviously writing back as original). I personally think this should be ok for handling most use-cases initially, and the code is structured in a way we could easily add more features without breaking changes in the future.

### Testing

Apart from unit, I added one happy path test as acceptance.

Manually, I tested with both SerDes, both gzip and uncompressed, found many bugs on the way and fixed all, adding relevant unit tests. 
For lake size, I went up to 100GB for now, which worked fine. In that test, the individual file size was about ~25MB, up to ~100MB uncompressed. It took about ~1h for 310 partitions and 50 fargate tasks. One thing I noticed, is that Athena seems to take longer than Parquet, with very big waiting time spent on queing - but all works fine consistently.

I also made a separate manual test with bigger files, ~800MB uncompressed, all good and similar patterns in terms of cpu, linear memory usage increase.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed


backend/ecs_tasks/delete_files/json_handler.py
```diff
@@ -0,0 +1,66 @@
...
+    for found_key in obj.keys():
+        if key.lower() == found_key.lower():
+            return found_key
+
+
+def delete_matches_from_json_file(input_file, to_delete, compressed=False):
+    deleted_rows = 0
+    with pa.BufferOutputStream() as out_stream:
+        input_file, writer = initialize(input_file, out_stream, compressed)
+        content = input_file.read().decode("utf-8")
```
HieronymusLex, line 40: do we need to state UTF8 encoding as a limitation somewhere in the docs or is that overkill?

backend/ecs_tasks/delete_files/json_handler.py
```diff
@@ -0,0 +1,66 @@
...
+        lines = content.split("\n")
+        if lines[-1] == "":
+            lines.pop()
+        total_rows = len(lines)
+        for line in lines:
+            parsed = json.loads(line)
```
HieronymusLex, line 46: one small issue here is that if a row in the JSON file contains invalid JSON, a JSONDecodeError will be thrown (subclass of ValueError) which will end up being caught in the ValueError block in `main.py` and the message might be confusing because it'll say something like "Unprocessable message: some error about JSON structure" however you can't tell from that error whether it's a problem with the message which was placed on the queue, or the file itself. You could tellby looking at the container logs since you could see from other log statements that the queue message was read OK, but on the error event it wouldn't be clear that the data was messed up. Perhaps we should handle decode errors here an rethrow a value error but with a more helpful error message

backend/ecs_tasks/delete_files/main.py
```diff
@@ -98,10 +108,9 @@ def execute(queue_url, message_body, receipt_handle):
         with s3.open(object_path, "rb") as f:
             source_version = f.version_id
             logger.info("Using object version %s as source", source_version)
-            infile = load_parquet(f)
             # Write new file in-memory
-            logger.info("Generating new parquet file without matches")
-            out_sink, stats = delete_matches_from_file(infile, cols)
+            compressed = object_path.endswith(".gz")
```
HieronymusLex, line 112: we should mention in the docs that we determine compression based on file extension (similar to the Athena docs https://docs.aws.amazon.com/athena/latest/ug/compression-formats.html)

frontend/src/components/pages/NewDataMapper.js
```diff
@@ -104,13 +104,15 @@ export default ({ gateway, goToDataMappers }) => {
     if (isFormValid) {
       setFormState("loading");
       try {
+        const format = selectedTable.format;
```
HieronymusLex, line 107: nit: don't need declare this const since it's only used once

HieronymusLex (changes requested):
Really nice work! only a couple of minor comments. Can you also run a test with large individual files e.g. 1GB to ensure that it performs as expected with those datasets

frontend/src/components/pages/NewDataMapper.js
matteofigus: done

backend/ecs_tasks/delete_files/json_handler.py
matteofigus: Done - I tried to construct a meaningful sentence out of the error and added a test. 
In general, we have to consider that Athena should possibly blow if there's any malformed JSON as we don't allow the ignore option, so this is possibly a edge case that would be in need of being dealt with only if you put a new object between Find and Forget, but I think it's a good call, allows better troubleshooting!

backend/ecs_tasks/delete_files/json_handler.py
matteofigus: Not sure, sounds like overkill to me? I'm unopinionated - happy to add a note but I didn't see anything similar in Athena and similar tools

matteofigus:
> Really nice work! only a couple of minor comments. Can you also run a test with large individual files e.g. 1GB to ensure that it performs as expected with those datasets

Thanks for the review! I addressed your comments and I made a new test putting together a single file of about ~10GB, compressed to gzip which resulted ~1.8GB. I believe that to be quite extraordinary as size but it still worked fine, took about 20m and took about 80% of memory when at peak (which I guess it makes sense given the uncompressed size of input+ouput+gzipped content.

HieronymusLex (approved):

PR #173, opened 2020-07-03
More supported data types, docs update

matteofigus:
*Issue #, if available:*
We only support a number of data types. With this change, I updated the docs, plus I updated the frontend to make it easier to show what's allowed or not.

Note: Future PRs are coming for supporting more types.
I left out here particularly dates, decimal, and complex structures. One about struct in particular, will allow all the allowed types when nested inside structs. When that will happen, I'll update the docs too.

Screenshot 2020-07-03 at 15 52 16
Screenshot 2020-07-03 at 15 52 46

*PR Checklist:*

- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@
 ### Summary
 
 - [#173](https://github.com/awslabs/amazon-s3-find-and-forget/pull/173): Show
-  column types and hierarchy in the front-end during Data Mapper creation + add
+  column types and hierarchy in the front-end during Data Mapper creation
+- [#173](https://github.com/awslabs/amazon-s3-find-and-forget/pull/173): Add
   support for char, smallint, tinyint, double, float
 
 ## v0.5
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+## 0.6 (unreleased)
+
+### Summary
+
+- [#173](https://github.com/awslabs/amazon-s3-find-and-forget/pull/173): Show
+  column types and hierarchy in the front-end during Data Mapper creation + add
+  support for char, smallint, tinyint, double, float
+
 ## v0.5
 
 ### Summary
--- a/backend/lambdas/tasks/generate_queries.py
+++ b/backend/lambdas/tasks/generate_queries.py
@@ -114,9 +114,11 @@ def convert_to_col_type(val, col, table):
 
     col_type = column["Type"]
 
-    if col_type == "string" or col_type == "varchar":
+    if col_type in ["char", "string", "varchar"]:
         return str(val)
-    if col_type == "int" or col_type == "bigint":
+    if col_type in ["bigint", "int", "smallint", "tinyint"]:
         return int(val)
+    if col_type in ["double", "float"]:
+        return float(val)
 
     raise ValueError("Column {} is type {} which is not a supported column type for querying")
--- a/backend/lambdas/tasks/generate_queries.py
+++ b/backend/lambdas/tasks/generate_queries.py
@@ -114,11 +114,11 @@ def convert_to_col_type(val, col, table):
 
     col_type = column["Type"]
 
-    if col_type in ["char", "string", "varchar"]:
+    if col_type in ("char", "string", "varchar"):
         return str(val)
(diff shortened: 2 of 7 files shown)


frontend/src/utils/glueSerializer.js
```diff
@@ -1,6 +1,89 @@
...
+    "double",
+    "float",
+    "int",
+    "smallint",
+    "string",
+    "tinyint",
+    "varchar"
+  ];
+
+  const columnMapper = c => {
```
HieronymusLex, line 19: I think we could probably do with a test for this? also it's quite hard to follow what's happening and some tests would help with that

frontend/src/utils/glueSerializer.js
```diff
@@ -1,6 +1,89 @@
...
+        const isStruct = rest.startsWith(STRUCTPREFIX);
+        const isStructArray = rest.startsWith(ARRAYSTRUCTPREFIX);
+
+        if (isStruct || isStructArray) {
+          const prefix = isStruct ? STRUCTPREFIX : ARRAYSTRUCTPREFIX;
+          let nOpenedTags = isStruct ? 1 : 2;
+          let endIndex = -1;
+          const structSchema = [...rest.substr(prefix.length)];
+          structSchema.some((c, i) => {
+            if (c === "<") nOpenedTags++;
```
HieronymusLex, line 60: not too sure about callback passed to `some` mutating vars from the outer scope. doesn't seem like a good approach from an functional perspective (not that were doing FP in this project!!!)

frontend/src/utils/glueSerializer.js
HieronymusLex: can we break this up a bit to make it more readable?

HieronymusLex:
@matteofigus I think we should be updating the changelog as we go e.g. add a new section to the changelog for the next version (marked as draft/pending) then we add to it as we merge PRs - thoughts?

frontend/src/utils/glueSerializer.js
matteofigus: done

matteofigus: done

frontend/src/utils/glueSerializer.js
matteofigus: I think you're right. Moved to for, which makes it actually more readable

backend/lambdas/tasks/generate_queries.py
```diff
@@ -114,9 +114,11 @@ def convert_to_col_type(val, col, table):
 
     col_type = column["Type"]
 
-    if col_type == "string" or col_type == "varchar":
+    if col_type in ["char", "string", "varchar"]:
```
ctd, line 117:
```suggestion
    if col_type in ("char", "string", "varchar"):
```

Nitpick. Tuples are more appropriate here.

backend/lambdas/tasks/generate_queries.py
```diff
@@ -114,9 +114,11 @@ def convert_to_col_type(val, col, table):
...
         return str(val)
-    if col_type == "int" or col_type == "bigint":
+    if col_type in ["bigint", "int", "smallint", "tinyint"]:
```
ctd, line 119:
```suggestion
    if col_type in ("bigint", "int", "smallint", "tinyint"):
```

Nitpick. Tuples are more appropriate here.

backend/lambdas/tasks/generate_queries.py
```diff
@@ -114,9 +114,11 @@ def convert_to_col_type(val, col, table):
...
         return int(val)
+    if col_type in ["double", "float"]:
```
ctd, line 121:
```suggestion
    if col_type in ("double", "float"):
```

Nitpick. Tuples are more appropriate here.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Change Log
 
+## 0.6 (unreleased)
+
+### Summary
+
+- [#173](https://github.com/awslabs/amazon-s3-find-and-forget/pull/173): Show
```
ctd, line 7: Since there's two changes (the interface change, and the additional types change), what about listing them as two points in the summary list.

matteofigus: Ok done

PR #174, opened 2020-07-09
Support for types nested in struct

matteofigus:
*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed


backend/ecs_tasks/delete_files/parquet.py
```diff
@@ -15,10 +15,40 @@ def get_row_count(df):
...
     for column in to_delete:
-        df = df[~df[column["Column"]].isin(column["MatchIds"])]
-    return df
+        indexes = df[column["Column"]].isin(column["MatchIds"])
+        indexes_to_delete.append(indexes)
+        df = df[~indexes]
+    if needs_flattened_columns:
+        df = table.to_pandas()
+        for indexes in indexes_to_delete:
+            df=df[~indexes]
```
ctd, line 48: For my benefit, could you explain what's happening in this block?

It looks like we're throwing away the value of `df` that's been built up during the previous loop then doing another similar loop. If that's the case it seems like there's room for optimisation?

But there could be some nuance I'm missing.

matteofigus: Ok. I put a general comment in the `needs_flattening` method but I can put it back here referencing direct lines of code. The problem is that when you flatten the DF that gets converted from 
```
field1|field2|field3
foo|{x=1,y=2}|bar
```
to
```
field1|field2.x|field2.y|bar
foo|1|2|bar
```
and if I write as flattened, that won't be the same schema as when I read that. That's why I use the flattened for the find, but then I re-iterate writing on the same indexes.

In terms of the specific line `df = df[df~indexes]` on the flattened DF, the reason I do it is that if I have multiple columns, that would guarantee that while re-iterating as a part of the second for loop, I would delete the same indexes. Do you think we can optimise further there? I'm thinking maybe I can just merge indexes in the `indexes_to_delete` (instead of keeping it as array) and avoid executing the first `df = df[df~indexes]` command...

ctd: Ah, ok. Thanks for the explanation.

I think this is fine as it is - I don't see obvious gains to trying to optimise it further.

PR #186, opened 2020-08-13
Fix Sanitiser

HieronymusLex:
*Issue #, if available:*
n/a
*Description of changes:*
Fixes the Forget ECS Task match ID sanitiser to handle integers

*PR Checklist:*

- [x] Changelog updated
- [X] Unit tests (and integration tests if applicable) provided
- [x] All tests pass
- [X] Pre-commit checks pass
- [X] Debugging code removed
- [x] If releasing a new version, have you bumped the version in the main CFN template?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #188, opened 2020-08-17
Unlimited Job's Deletion Queue

ormahler:
*Issue #, if available:*
Multiple columns still not supported
*Description of changes:*
 Save Jobs Deletion Queue to S3 and pass it by reference through the components  
*PR Checklist:*

- [ ] Changelog updated
- [ ] Unit tests (and integration tests if applicable) provided
- [ ] All tests pass
- [ ] Pre-commit checks pass
- [ ] Debugging code removed


HieronymusLex:
@ormahler Thanks for the PR! We're looking into the changes introduced here at the moment but we may break this PR down into several smaller one's. We'll provide updates as we make progress

ormahler:
Hi Nick,

Thanks for the follow up.
Seems right,  please just be aware that there are changes in different
components which are dependent on one another.
I would gladly meet you to discuss it.

Regards,
Or

backend/ecs_tasks/delete_files/main.py
```diff
@@ -104,26 +118,30 @@ def execute(queue_url, message_body, receipt_handle):
             logger.info("Generating new parquet file without matches")
             out_sink, stats = delete_matches_from_file(infile, cols)
         if stats["DeletedRows"] == 0:
-            raise ValueError(
-                "The object {} was processed successfully but no rows required deletion".format(
-                    object_path
+            if all_files:
```
matteofigus, line 121: Looks like you are adding a setting to skip throwing for no deletions in file, can I know the reason for that?

backend/ecs_tasks/delete_files/main.py
```diff
@@ -63,9 +66,16 @@ def handle_error(
             logger.error("Unable to change message visibility: %s", str(e))
 
 
+def parse_msg_body_and_revert_last(message_body, client):
+    body = json.loads(message_body)
+    object_path = body.get("Object")
+    input_bucket, input_key = parse_s3_url(object_path)
+    revert_last(client, input_bucket, input_key)
```
matteofigus, line 73: I can't see any usage of this function?

backend/lambdas/jobs/stats_updater.py
```diff
@@ -30,82 +33,86 @@ def _aggregate_stats(events):
...
+                "TotalQueryScannedInBytes": event_data.get("Statistics", {}).get("DataScannedInBytes", 0),
+                "TotalQueryTimeInMillis": event_data.get("Statistics", {}).get("EngineExecutionTimeInMillis", 0),
+            })
+        if event_name in ["ObjectUpdated", "ObjectUpdateFailed", "ObjectRollbackFailed"]:
+            stats += Counter({
+                "TotalObjectUpdatedCount": 1 if event_name == "ObjectUpdated" else 0,
+                "TotalObjectUpdateFailedCount": 1 if event_name == "ObjectUpdateFailed" else 0,
+                "TotalObjectRollbackFailedCount": 1 if event_name == "ObjectRollbackFailed" else 0,
+            })
+            if event_name == "ObjectUpdateFailed":
```
matteofigus, line 49: ObjectUpdateFailed is an event triggered for a single object update fail, and it means there was no write - so, nothing to revert as deleting the previous version may mean deleting the only version there.

The revert is done from the task currently - why moving it to lambda? I think I'm a bit lost here.

backend/lambdas/jobs/stats_updater.py
```diff
@@ -30,82 +33,86 @@ def _aggregate_stats(events):
...
+                    data_access_role_arn = event_data["Message"].get("RoleArn", getenv("DataAccessRoleArn"))
+                    session = get_session(data_access_role_arn)
+                    client = session.client("s3")
+                    s3_object = event_data["Message"]["Object"]
+                    input_bucket, input_key = parse_s3_url(s3_object)
+                    revert_last(client, input_bucket, input_key)
+                except Exception:
+                    logger.exception("Unable to revert last")
+                    try:
+                        purge_queue(deletion_queue_url)
```
matteofigus, line 61: Why purging the queue in case of error? I'm curious. You prefer to quit as soon as you have an error and restart a job rather than keep trying for other objects first?

backend/lambdas/tasks/generate_queries.py
```diff
@@ -15,115 +16,111 @@
...
-    "bigint",
-    "char",
-    "double",
-    "float",
-    "int",
-    "smallint",
-    "string",
-    "tinyint",
-    "varchar",
-]
```
matteofigus, line 36: This stuff (and more at the bottom of the file) is all needed when using column identifiers nested inside structs - I guess you removed it because you didn't need it, or was it just a merge conflict?

matteofigus (commented):
Thanks for submitting your changes. I added a couple of questions as I start planning the work involved

backend/lambdas/jobs/stats_updater.py
ormahler: From our experience, there were many cases in which ObjectUpdateFailed event indicated that the file was corrupted, causing both Athena and spark which are querying this dataset to fail.
therefore when ObjectUpdateFailed occurs, we try to revert last version.
revert_last() function makes 2 validations:
    1) checks that there are prior versions for this file.
    2) checks that the previous version is not a 'delete marker'.
only if these two conditions are met the revert happens. 
this way we assure we no file will be deleted.
* as the comment below indicates this flow is only usable when deleteOldVersion flag is set to True.

We encountered some cases in which ObjectUpdateFailed was caused by SIGINT/SIGTERM and the deletion task wasn't able to complete the revert before it was down, therefore we moved it to the stream processor.

backend/lambdas/jobs/stats_updater.py
ormahler: As described in the reply above, ObjectUpdateFailed may indicate file corruption,
hence in case the revert failed we want to stop the job and check if the data is corrupted.
if it is, we need to locate the corrupted file and revert it manually.
This state is not expected to happen therefore assuming it did we rather stop and find the cause in order to prevent the corruptions of further files.

backend/lambdas/jobs/stats_updater.py
matteofigus: Ok, just to clarify, the file was corrupted before already, or by the solution during a Deletion Job?

ormahler: caused by the solution

matteofigus: Ok, this is pretty interesting, it would be great if we could find out what the issue is and reproduce to create a fix.

Do you have any details about what the corrupted objects look like? For instance, 
* do you still have a full log for the `ObjectUpdateFailed` job event?
* do you know if before/after the objects had similar filesize?
* any correlation to specific partitions or it would happen anywhere?
* could it be related to this fixes we did a couple of weeks ago? #193 #186

ormahler: This is the Athena error message we get:
  ```
Your query has the following error(s):
HIVE_CANNOT_OPEN_SPLIT: Error opening Hive split s3://neura-s3f2-test/testing/neura-emr-raw- 
data/year=2020/month=01/day=20/part-00616-b59d4d11-c5bb-433e-92cb-c78623cfa3f0-c000.snappy.parquet (offset=0, length=2480536): null
```
 
Even further interesting is that this error occurs in two scenarios:
    1) Alternately during the forget phase (for successful jobs, when the job ends it stops occurring)
    2) When there is a corrupted file in the dataset as described in the previous reply.

Therefore, we assume that there is a fraction of time during the file rewrite in which the file is not yet completed but is already a part of the dataset. then, if a forget task stops unexpectedly during this small phase (due to a SIGINT/SIGTERM for instance) 
the file remains corrupted in the dataset and the error keeps occurring.

you can try replicate scenario (1) by starting a deletion job and keep to querying the dataset until this error arrives.
I used:
```select * 
FROM [table_name]
where [match_key] in (...)
```
You can try and replicate scenario (2) by starting a deletion job over a dataset with rather large files (250 MB on our case) and to set fargate's Task memory to the smallest available value- 0.5GB  

I did save some logs locally somewhere, I will try to find them and will let you know.

matteofigus: Thanks for the details!

I was able to reproduce the error for the first scenario, and I think I have an explanation. Presto, running under Athena, inherits S3 consistency model and in some scenarios makes multiple calls to S3 to fetch the list of files of a given partition, and then to get the content, for instance getting the metadata, and then making multiple calls to fetch different row groups. It could occur that exactly when we put the new version, there is a timeframe where Presto could get a mix of old and new content. In the scenario of reading the metadata from the old file and then reading the rowgroups from new, the bytes wouldn't match up. This is an example error in my scenario: `HIVE_CANNOT_OPEN_SPLIT: Error opening Hive split s3://matteo-tests/cr_basic/product_category=Apparel/part-00000-495c48e6-96d6-4650-aa65-3c36a3516ddd.c000.snappy.parquet (offset=0, length=33554432): can not read class parquet.format.FileMetaData: null` - immediately after re-running the query, it was fine. Now...if you have a high number of fargate tasks and running a query to all partitions, I guess the probability of this error occurring may increase significantly, because while on a specific moment you may be having an issue with an object, a moment later, while the issue could be mitigated for the current one, you may be having it with another one. That would explain why after the job runs all is good. In general, I don't know if I have a solution to this problem, apart from recommending running the deletion job in a moment where read users wouldn't be impacted, or obviously increase the tasks concurrency to reduce the timeframe where this is issue is most likely to happen. 
Do you think this theory may be valid for the scenario n.1?
In this scenario, I think that I wouldn't probably recommend a rollback as it would be completely ineffective (it would just make Presto to suffer from the same issue another time, without de-facto solving the issue).

In regards of the scenario (2) - data staying corrupted after the job ends, I think it's a separate issue, and I'm still trying to reproduce.

ormahler: Sounds very reasonable, and indeed we don't rollback in scenario (1)
 
I emailed you a specific job summary and overview which might be relevant.

PR #191, opened 2020-08-19
Integrate VPC template into main template

HieronymusLex:
*Issue #, if available:*
N/A
*Description of changes:*
- Makes VPC details optional
- Deploys solution provided VPC custom VPC details not provided
- Remove VPC template from published files

*PR Checklist:*

- [x] Changelog updated
- [X] Unit tests (and integration tests if applicable) provided
- [X] All tests pass
- [X] Pre-commit checks pass
- [X] Debugging code removed
- [X] If releasing a new version, have you bumped the version in the main CFN template?


cfn-publish.config
```diff
@@ -1,5 +1,5 @@
 bucket_name_prefix="solution-builders"
 acl="public-read"
 extra_files=build.zip
-templates="templates/template.yaml templates/vpc.yaml templates/role.yaml"
+templates="templates/template.yaml templates/role.yaml"
```
HieronymusLex, line 4: I think it makes sense to stop publishing the VPC template since it's now "internal" to the solution but I'm open to suggestions

templates/template.yaml
```diff
@@ -300,6 +311,14 @@ Resources:
...
         JobTableStreamArn: !GetAtt DDBStack.Outputs.JobTableStreamArn
         StateMachineArn: !GetAtt StateMachineStack.Outputs.StateMachineArn
+  VpcStack:
+    Type: AWS::CloudFormation::Stack
+    Condition: DeployVpc
+    Properties:
+      TemplateURL: ./vpc.yaml
+      Parameters:
+        FlowLogsGroup: !Ref FlowLogsGroup
+        FlowLogsRoleArn: !Ref FlowLogsRoleArn
```
matteofigus, line 321: Should we allow passing down the other 2 params too? (the IP blocks to be used). Probably explaining those defaults will be ignored if you specify subnet/sec groups for existing VPC.

cfn-publish.config
matteofigus: I Agree.

templates/template.yaml
```diff
@@ -107,12 +115,15 @@ Parameters:
   VpcSecurityGroups:
     Description: List of security groups to apply to Fargate tasks
```
matteofigus, line 116: Should we specify here in the description that by leaving blank, a new VPC will be created? Just thinking of who doesn't read the docs

templates/template.yaml
```diff
@@ -107,12 +115,15 @@ Parameters:
...
     Type: CommaDelimitedList
+    Default: ""
   VpcSubnets:
     Description: List of subnets to deploy Fargate tasks in
```
matteofigus, line 120: Should we specify here in the description that by leaving blank, a new VPC will be created? Just thinking of who doesn't read the docs

matteofigus (changes requested):
I'm in favour of the proposed approach because it simplifies and generally is capable of handling the scenarios. I added only a couple of comments related to the CFN implementation.

templates/template.yaml
HieronymusLex: I wasn’t sure on that one. In my mind if you’re customising those, you may as well be using your own VPC?

matteofigus: The problem is that maybe you want the VPC but those CIDRs are already used? I guess an escape would be to deploy the VPC separately if that would be the case, but I think we also want to remove the links to that 😢

ctd (commented):
I feel really uneasy about presuming that a user will want the VPC stack launched based on an omitted parameter. I think it could cause unpleasant astonishment for users.

If we're going to have a single do-all-the-things template, I think there should be an explicit parameter for customers to make the decision if they want the VPC stack deployed. It should then also validate that the combination of parameters supplied are valid - e.g. requesting the provided VPC, then providing a custom Security Group or Subnet should result in an error.

templates/template.yaml
HieronymusLex: CIDR blocks already being used is only a problem if you can't have VPCs with overlapping IPv4 CIDR block (e.g. you intend to peer the S3F2 VPC?) but at that point you should be using your own VPC rather than the solution provided one because you'll be amending a CFN managed resource. so I still don't see a use case where you need to be able to customise all the params but still use the solution provided VPC rather than your own existing one?

HieronymusLex:
@ctd leaving parameters blank to enable/disable features is consistent with the approach we take for access logs but I suppose this is slightly different in that the default configuration has additional costs (though a properly configured VPC is a hard dependency for the solution to work). 

I think adding add in a true/false parameter to determine if we want the solution provided VPC makes things more explicit which I'm in favour of where there's costs involved. However, we can't do the suggested validation without something like a macro which I'm against adding in. How about:
- add new param `DeployVPC`
- if `DeployVpc` == true - deploy the VPC. no additional validation needed here
- if `DeployVpc` == false - don't deploy the VPC. validation remains as-is i.e. if you specify and invalid subnet ID or security group ID, the deletion task stack will fail to deploy. trade-off here is that we need to leave these params as CommaDelimitedList types (as-is today) rather List<AWS::EC2::Subnet::Id> since they need to be optional

matteofigus (approved):

docs/USER_GUIDE.md
```diff
@@ -45,20 +45,17 @@ Endpoints]:
...
-This template will create a VPC with private subnets and all the relevant VPC
-Endpoints required by the Amazon S3 Find and Forget solution. To deploy this
-template, use the VPC Template "Deploy to AWS button" in
-[Deploying the Solution](#deploying-the-solution) then follow steps 5-9. The
-**Outputs** tab will contain the subnets and security group IDs to use as inputs
-for the main stack.
+If you do not have an existing VPC you wish to use, a solution provided VPC can
+be created by setting _DeployVpc_ to true when deploying the solution. The
+solution provided VPC only has private subnets and has all the relevant VPC
+Endpoints required by the Amazon S3 Find and Forget solution.
```
ctd, line 51:
```suggestion
By default the CloudFormation template will create a new VPC that has been
purpose-built for the solution. The VPC includes VPC endpoints for the
aforementioned services, and does not provision internet connectivity.

You can use the provided VPC to operate the solution with no further
customisations. However, if you have more complex requirements it is
recommended to use an existing VPC as described in the following section.
```

I think it makes sense to position the provided VPC as the new default -- that unless you want/need to BYO, just use it and you're set.

docs/USER_GUIDE.md
```diff
@@ -45,20 +45,17 @@ Endpoints]:
...
 
 #### Using an Existing VPC
 
 If you wish to use an existing VPC in your account with the Amazon S3 Find and
```
ctd, line 55: I suggest replacing this paragraph with the following to provide better guidance on why you would want to use your own VPC, and elaborate on the parameters that need to be provided:
```suggestion
Amazon S3 Find and Forget can also be used in an existing VPC. You may
want to do this if you have requirements that aren't met by using the VPC
provided with the solution.

To use an existing VPC, set the `DeployVpc` parameter to `false` when
launching the solution CloudFormation stack. You must also specify the
subnet and security groups that the Fargate tasks will use by setting the
`VpcSubnets` and `VpcSecurityGroups` parameters respectively.

The subnets and security groups that you specify must allow the tasks to
connect to the aforementioned AWS services.
```

ctd (changes requested):
I'm happier with the implementation now - it's much clearer with the boolean toggle.

Just a few suggested improvements for the docs.

ctd:
Looking good. Is there a plan for how we will release/communicate this change? 

Just updating an existing stack to the new template probably won't do what users expect it to.

docs/UPGRADE_GUIDE.md
```diff
@@ -0,0 +1,15 @@
+# Upgrade Guide
+
+## Migrating from <=v0.8 to 0.9
+
+The default behaviour of the solution has been changed in v0.9 to deploy and use
+the VPC that has been purpose-built for the solution as a nested stack.
```
ctd, line 6:
```suggestion
The default behaviour of the solution has been changed in v0.9 to deploy and use
a purpose-built VPC when creating the solution CloudFormation stack.
```

docs/UPGRADE_GUIDE.md
```diff
@@ -0,0 +1,15 @@
...
+
+If you have deployed the standalone VPC stack provided in previous versions, you
+should should set `DeployVpc` to **true** when upgrading to v0.9 and input the
+same values for the `FlowLogsGroup` and `FlowLogsRoleArn`. parameters that were
```
ctd, line 10:
```suggestion
same values for the `FlowLogsGroup` and `FlowLogsRoleArn` parameters that were
```

docs/UPGRADE_GUIDE.md
```diff
@@ -0,0 +1,15 @@
...
+used when deploying the standalone VPC stack. After the deployment of v0.9 is
+complete, you may delete the old VPC stack.
```
ctd, line 12:
```suggestion
complete, you should delete the old VPC stack.
```

RFC2119 pedanticism.

docs/UPGRADE_GUIDE.md
```diff
@@ -0,0 +1,15 @@
...
+
+If you wish to continue using an existing VPC, you must set `DeployVpc` to
+**false** when upgrading to v0.9.
```
ctd, line 15:
```suggestion
To continue using an existing VPC, you must set `DeployVpc` to
**false** when upgrading to v0.9.
```

ctd (commented):
few small changes, otherwise LGTM

ctd (approved):

PR #200, opened 2020-09-17
Batch handling for deletion queue items

matteofigus:
*Description of changes:*
Adds a batch endpoint to allow populating the queue quicker.
I'm not entirely sure of the API here, we already have a `PATCH /queue` for inserting an item, I thought we could just have a breaking change or keep the API of `queue/matches` compatible with the DELETE one (which uses that). So at least this is not breaking change.

One other thing I'm not sure is if we should set an arbitrary limit given we are subject to API Gateway payload size and response size, for instance 100 items or so?

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## v0.11 (unreleased)
 
 - [#200](https://github.com/awslabs/amazon-s3-find-and-forget/pull/200): Add API
-  Endpoint for adding deletion queue items in batch
+  Endpoint for adding deletion queue items in batch - deprecates PATCH /v1/queue
 - [#170](https://github.com/awslabs/amazon-s3-find-and-forget/pull/170): JSON
   support
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## v0.11 (unreleased)
 
+- [#200](https://github.com/awslabs/amazon-s3-find-and-forget/pull/200): Add API
+  Endpoint for adding deletion queue items in batch
 - [#170](https://github.com/awslabs/amazon-s3-find-and-forget/pull/170): JSON
   support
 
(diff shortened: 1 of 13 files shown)


HieronymusLex:
I don't think /queue/matches is quite right in RESTful terms but tbh I don't think it matters a great deal. If we're sticking with 2 endpoints, I'd at least like the code to be shared between the 2 

```
def enqueue_handler(event, context):
    body = event["body"]
    match_id = body["MatchId"]
    data_mappers = body.get("DataMappers", [])
    item = enqueue_matches([body])[0]

    return {"statusCode": 201, "body": json.dumps(item, cls=DecimalEncoder)}


@with_logging
@add_cors_headers
@json_body_loader
@catch_errors
def enqueue_batch_handler(event, context):
    body = event["body"]
    matches = body["Matches"]
    items = enqueue_matches(matches)
    return {
        "statusCode": 201,
        "body": json.dumps({"Matches": items}, cls=DecimalEncoder),
    }

def enqueue_items(matches):
    items = []
    with deletion_queue_table.batch_writer() as batch:
        for match in matches:
            match_id = match["MatchId"]
            data_mappers = match.get("DataMappers", [])
            item = {
                "DeletionQueueItemId": str(uuid.uuid4()),
                "MatchId": match_id,
                "CreatedAt": utc_timestamp(),
                "DataMappers": data_mappers,
                "CreatedBy": get_user_info(event),
            }
            batch.put_item(Item=item)
            items.append(item)
    return items
```

matteofigus:
> I think we should remove or at the very least deprecate the enqueue item endpoint since it's being superseded by this newly introduced endpoint

Cool, I refactored the code and added a deprecation notice in the api spec. While the deprecated flag seems supported by OpenAPI I didn't see anything changing on the generated docs, so I updated the description of the method to clarify it's deprecated and mentioned it in the Changelog. I think we can keep it alive for another couple of versions, and then completely remove it.

HieronymusLex (approved):

PR #202, opened 2020-09-22
Fix query failures due to incorrect partition casting

matteofigus:
*Description of changes*
Adds handling for non-string partition types such as int. I opted for keeping the strategy consistent with what we do with matches: we get the partition type from glue, then we cast it during the generate phase, so that the execute inherits the values already casted. 

Credits: I took some code snippets from @ormahler's #188 PR 

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed

### DIFF
--- a/tests/acceptance/conftest.py
+++ b/tests/acceptance/conftest.py
@@ -256,6 +256,7 @@ def glue_table_factory(dummy_lake, glue_client, glue_columns):
         prefix="prefix",
         partition_keys=[],
         partitions=[],
+        partition_key_types="string",
     ):
         glue_client.create_database(DatabaseInput={"Name": database})
         input_format = (
@@ -292,7 +293,7 @@ def glue_table_factory(dummy_lake, glue_client, glue_columns):
                     "StoredAsSubDirectories": False,
                 },
                 "PartitionKeys": [
-                    {"Name": pk, "Type": "string",} for pk in partition_keys
+                    {"Name": pk, "Type": partition_key_types} for pk in partition_keys
                 ],
                 "Parameters": {"EXTERNAL": "TRUE"},
             },
@@ -353,6 +354,7 @@ def glue_data_mapper_factory(
         role_arn=None,
         delete_old_versions=False,
         column_identifiers=["customer_id"],
+        partition_key_types="string",
     ):
         item = {
             "DataMapperId": data_mapper_id,
@@ -377,6 +379,7 @@ def glue_data_mapper_factory(
             table=table,
             partition_keys=partition_keys,
             partitions=partitions,
+            partition_key_types=partition_key_types,
         )
 
         items.append(item)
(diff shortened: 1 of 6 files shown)


tests/unit/tasks/test_generate_queries.py
```diff
@@ -622,9 +662,17 @@ def test_it_throws_for_unconvertable_matches(self):
...
+        with pytest.raises(ValueError) as e:
             get_inner_children("struct<name:string", "struct<", ">")
+        assert e.value.args[0] == "Column schema is not valid"
+
+    def test_it_throws_for_invalid_schema_for_nested_children(self):
+        with pytest.raises(ValueError) as e:
+            get_nested_children(
+                "struct<name:string,age:int,s:struct<n:int>,b:string", "struct"
+            )
+        assert e.value.args[0] == "Column schema is not valid"
```
matteofigus, line 675: This is slightly unrelated, there was no coverage for it so I just added those to make sure that scenario is handled

tests/unit/tasks/test_execute_query.py
```diff
@@ -68,6 +68,42 @@ def test_it_generates_query_with_partition():
...
+            "PartitionKeys": [{"Key": "product_category", "Value": "Books"}],
+        }
+    )
+
+    assert (
+        escape_resp(resp) == 'SELECT DISTINCT "$path" '
+        'FROM "amazonreviews"."amazon_reviews_parquet" '
+        'WHERE ("customer_id" in (123456, 456789)) '
+        "AND \"product_category\" = 'Books'"
+    )
```
matteofigus, line 86: Slightly unrelated but there was no test on the execute query for int matches so I added it just for clarity

tests/unit/tasks/test_generate_queries.py
```diff
@@ -690,7 +738,9 @@ def partition_stub(values, columns, table_name="test_table"):
     }
 
 
-def table_stub(columns, partition_keys, table_name="test_table"):
+def table_stub(
+    columns, partition_keys, table_name="test_table", partition_key_types="string"
```
HieronymusLex, line 742: nit: types? shouldn't this be type if the value isn't a list?

tests/acceptance/conftest.py
```diff
@@ -256,6 +256,7 @@ def factory(
         prefix="prefix",
         partition_keys=[],
         partitions=[],
+        partition_key_types="string",
```
HieronymusLex, line 259: nit: types? shouldn't this be type if the value isn't a list?

HieronymusLex (approved):
couple of very minor comments. LGTM though

tests/unit/tasks/test_generate_queries.py
matteofigus: you're right, renamed to partition_keys_type

HieronymusLex (approved):

PR #208, opened 2020-11-03
Remove ambiguity between DataMappers Create/Modify/Put

matteofigus:
*Description of changes:*

At the moment the PUT data mapper is advertised as Create in API/Frontend and I think this generates confusion. I renamed the "Create" in the backend to be PUT in order to be consistent with API behaviour. I also found and fixed a bug involving doing a PUT for an existing DM that had the same location being prevented from being modified.

In the frontend, I think it makes sense to keep the naming and behaviour as "create", so I decided to add a check to prevent adding a new data mapper with the same ID both on validation phase (by fetching all DMs on page load and then checking synchronously against that) and submit phase (checking one last time to reduce probability of a race condition).

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed

### DIFF
--- a/frontend/src/components/pages/NewDataMapper.js
+++ b/frontend/src/components/pages/NewDataMapper.js
@@ -63,6 +63,7 @@ export default ({ gateway, goToDataMappers }) => {
 
   const [dataMapperId, setDataMapperId] = useState(undefined);
   const [errorDetails, setErrorDetails] = useState(undefined);
+  const [existingDataMapperIds, setExistingDataMapperIds] = useState([]);
   const [formState, setFormState] = useState("loading");
   const [glueData, setGlueData] = useState(undefined);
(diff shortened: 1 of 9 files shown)


frontend/src/components/pages/NewDataMapper.js
```diff
@@ -235,6 +250,11 @@ export default ({ gateway, goToDataMappers }) => {
                   onChange={(e) => setDataMapperId(e.target.value)}
                   {...validationAttributes(isDataMapperIdValid)}
                 />
+                {submitClicked && overlappingDataMappers && (
+                  <span className="form-error">
```
HieronymusLex, line 254: nit: should probably be a reusable component

HieronymusLex (approved):
LGTM

PR #226, opened 2020-12-11
Support for composite match IDs

### LINKED ISSUE #218: Support for composite match ids with multi-column data mappers
It is currently possible to select multiple columns as identifiers for data mappers.

When adding a match id to the queue, then, that's mapped to all columns, and there is no possibility to match particular values to particular columns.

Let's take as example a table:

```
customer_id | first_name | last_name
```

If we have scenarios of a an anonymous user (for which customer_id is `null`) and we want to match against a tuple of first_name/last_name, it is not currently possible.

With this feature we want to ensure it is possible to add matches like:
* value=12345 to be matched against `customer_id`
* value=John to be matched against `first_name` *AND* value=Doe to be matched against `last_name`

matteofigus:
*Issue #, if available:*
closes #218 

# Description

This PR is for adding support to Composite MatchIds to S3F2.
Composite Match Ids are supposed to help with scenarios that we have seen with many customers related to having rows that cannot have a direct match with a particular column. For instance, given a similar dataset:

|customer_id|first_name|last_name|
|-|-|-|
|12345|John|Doe|
|0|Jane|Doe| 

Where we have both registered and anonymous users. In one scenario, it could be possible to identify a customer by ID, in another we may need to act upon multiple fields.

With composite matches, we can now select all the 3 fields as column identifiers, and then we can add a composite match to identify a user by providing values for particular fields.

## API and Backwards compatibility

I decided to introduce a new field identifying the Match in the API called `Type`=`Simple|Composite`. The API is backwards compatible as creating a match (single or batch) using the old API will now create a match of Type=`Simple` as default. Old matches in Dynamo that don't contain the Type will still be backward compatible as covered by a acceptance test. To keep the API simple, I decided to keep the same endpoint for creating matches of simple or composite type. As a consequence, I decided that:
1) MatchId can now be a string or an array of `{Column, Value}` objects including the Column name and value of each value (min 1).
2) The `DataMappers` field stays unchanged but for composite I cannot allow more than 1 datamapper (as it maps to specific fields), so I decided to add extra server-side validation to check that the `DataMappers` array contains exactly 1 item if Type=`Composite`. We can change this in the future (for instance for datamappers that have similar schema or share column identifiers with the same names and types) but I decided this was the safest way to start for now.
3) The /queue API will return a Type on each queue item response, regardless of it being in Dynamo (deployed with a new version) or not (backward compatibile)

## Python code: Querying Athena

I wanted to identify I simple way to query Athena with hundreds of matches without creating GB big queries. An approach I came up with, is concatenating composite matches during query using an arbitrary token (`____`). Any OR statement is then grouped by matching strategy, and we obtain a query size that grows in complexity linearly compared to the simple match ids.  Example:

```
SELECT DISTINCT "$path"
FROM "amazonreviews"."amazon_reviews_parquet"
WHERE ("a"."b"."c" in ('a123456', 'b123456')
OR concat("user"."first_name", '____', "user"."last_name")
in ('John____Doe', 'Jane____Doe'))
```
As you can imagine, this is what inspired me to call the this mechanism `Composite` (therefore keeping the naming singular for the match Id despite being composed of multiple values).

## Python code: Anything else

Apart from querying athena (execute_query lambda) I think the rest is quite boring - most of the changes are for propagating this data from Find to Forget and I tried to translate any possible edge case to a unit test, and added pretty good number of acceptance too. But let me know if you think something can be simplified.

## Web UI

I tried to keep the UI simple as it could get hard to navigate this various validation flows. So initially I ask the user to pick between Simple and Composite, and I hide Composite if I know there isn't at least 1 Data Mapper with more than 1 column identifier. 
Screenshot 2020-12-15 at 18 58 58
The second step is showing only the Data Mappers with more than 1 column identifier:
Screenshot 2020-12-15 at 18 59 16
Then I allow to select which ones to provide values for, and I allow blank values.
Screenshot 2020-12-15 at 18 59 29

## Code Review

I know the code is a lot! For the review I recommend breaking it down to these units:

1) Review the terminology and docs. Also is "Composite" is a good name?
2) Review the API changes
3) Review the Python code and its tests
4) Review the Frontend code

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed


ctd (commented):
Changes look generally good to me! Thank you for the detailed change description and screenshots.

The composite approach is an elegant way of fitting it into the design. My only question is about the choice of key itself (commented on the line where you set the constant).

backend/ecs_tasks/delete_files/parquet_handler.py
```diff
@@ -51,15 +81,22 @@ def delete_from_table(table, to_delete):
     for column in to_delete:
```
HieronymusLex, line 81: nit: the little bit of repetition here suggests to me that `to_delete` should contain both normal and composite deletions, then we should determine the correct function call based on the deletion type e.g.

```
for column in to_delete:
        is_composite = # some way of finding out if it's a composite or simple
        get_row_indexes = get_row_indexes_to_delete if is_composite else get_row_indexes_to_delete_for_composite
        indexes = get_row_indexes(
            table, column["Columns"], column["MatchIds"]
        )
        table = table.filter(~indexes)
```

HieronymusLex: not a huge deal if it would be problematic to do it like this, but if it's possible we should consider it

frontend/src/components/FormRadioBinaryChoice.js
```diff
@@ -0,0 +1,30 @@
...
+  name,
+  onlyChoice = false,
+  onSelect,
+  selected,
+  title
+}) => (
+  <div
+    className={`radio-card ${selected && "selected"}`}
+    style={{
+      width: "49%",
```
HieronymusLex, line 16: nts what this is doing but it looks like something that could be better achieved with flexbox? difficult to say without knowing what the problem was but I'm always wary when I see things like 49%!

HieronymusLex:
a few minor comments but in general LGTM - great job!

backend/ecs_tasks/delete_files/json_handler.py
```diff
@@ -30,7 +30,25 @@ def find_key(key, obj):
...
 
-def delete_matches_from_json_file(input_file, to_delete, compressed=False):
+def get_value(key, obj):
+    """
+    Method to find a value given a nested key in an object. Example:
+    key="user.Id"
+    obj='{"user":{"id": 1234}}'
+    result=1234
+    """
+    record = obj
```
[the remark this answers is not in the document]
matteofigus: You're right, done.

backend/lambdas/tasks/execute_query.py
```diff
@@ -6,6 +6,8 @@
 
 client = boto3.client("athena")
 
+COMPOSITE_JOIN_TOKEN = "____"
```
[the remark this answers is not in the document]
matteofigus: Very good idea, done.

backend/lambdas/tasks/execute_query.py
```diff
@@ -43,19 +49,41 @@ def make_query(query_data):
...
     db = query_data["Database"]
     table = query_data["Table"]
     columns = query_data["Columns"]
+    composite_columns = query_data["CompositeColumns"]
     partitions = query_data.get("PartitionKeys", [])
 
     column_filters = ""
     for i, col in enumerate(columns):
         if i > 0:
             column_filters = column_filters + " OR "
```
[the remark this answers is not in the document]
matteofigus: You're right, I updated with `+=` + `itemgetter` to simplify.

backend/ecs_tasks/delete_files/parquet_handler.py
matteofigus: You're right. I went on extending the `Columns` object to contain a `Type` property for each match, and that allowed me to simplify the code a lot in most of the lambdas and the ecs task too! Thanks

frontend/src/components/FormRadioBinaryChoice.js
matteofigus: So you can see an example of this component here:
Screenshot 2020-12-17 at 02 34 21

To refactor given I have a similar binary radio choice later (all Data mappers vs specify Data Mappers), I created this component here `FormRadioBinaryChoice`. As you can see in the following line, I have 49% of width + 1% of margin right (that is what I already had in the previous iteration).

The only new thing here is adding this param `onlyChoice` where in case I add 51% because `Composite` is not an available option and I basically don't want this to happen:
Screenshot 2020-12-17 at 02 41 20

I'm not super familiar with flexbox but I can try to dig in if you think it could be a good improvement. 
What do you think?

matteofigus:
> Changes look generally good to me! Thank you for the detailed change description and screenshots.
> The composite approach is an elegant way of fitting it into the design. My only question is about the choice of key itself (commented on the line where you set the constant).

Thanks! I made amendments as suggested, good ideas.

> a few minor comments but in general LGTM - great job!

Thanks! I refactored the python code, probably worth a new look as I made a couple of changes to improve performance, hopefully it's still all looking quite pythonic. In regards of the frontend comment, I provided a little clarification, let me know if you think it's still worth investigating flexbox.

frontend/src/components/FormRadioBinaryChoice.js
HieronymusLex: ah ok I see what you mean. probably. a good candidate for something like bootstraps responsive grid system. I was worrying that when the screen is small it'll look a bit weird at 49%?

HieronymusLex:
> Thanks! I refactored the python code, probably worth a new look as I made a couple of changes to improve performance, hopefully it's still all looking quite pythonic

looks much better now I think. nice work

HieronymusLex (approved):

ctd (approved):

PR #236, opened 2021-02-05
Export API Gateway URL + Deletion Queue Stream ARN

matteofigus:
*Description of changes:*
This is to make it easier for customers to create applications that integrate with the API, and need to create a stream to verify queue item creation/deletion.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed

### DIFF
--- a/docs/USER_GUIDE.md
+++ b/docs/USER_GUIDE.md
@@ -15,6 +15,7 @@ Forget solution.
   - [Logging in for the first time](#logging-in-for-the-first-time)
   - [Managing users](#managing-users)
   - [Making authenticated API requests](#making-authenticated-api-requests)
+  - [Integrating the solution with other applications using CloudFormation stack outputs](#integrating-the-solution-with-other-applications-using-cloudformation-stack-outputs)
 - [Configuring Data Mappers](#configuring-data-mappers)
   - [AWS Lake Formation Configuration](#aws-lake-formation-configuration)
   - [Data Mapper Creation](#data-mapper-creation)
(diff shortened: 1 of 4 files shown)


docs/USER_GUIDE.md
```diff
@@ -318,6 +319,22 @@ requests using the AWS CLI:
...
+url) as input for another application. Alternatively, you can use some of the
+exported values for deploying separate stacks.
+
+After another stack imports an output value, you can't delete the stack that is
+exporting the output value or modify the exported output value. All of the
+imports must be removed before you can delete the exporting stack or modify the
+output value.
+
+Consult the [exporting stack output values] guide and review the differences
+between importing exported values vs using nested stacks.
```
ctd, line 336:
```suggestion
Applications deployed using AWS CloudFormation in the same AWS account and
region can integrate with Find and Forget by using CloudFormation output
values. You can use the solution stack as a nested stack to use its outputs
(such as the API URL) as inputs for another application. 

Some outputs are also available as exports. You can import these values to use
in your own CloudFormation stacks that you deploy following the Find and Forget
stack.

**Note for using exports:** After another stack imports an output value, you
can't delete the stack that is exporting the output value or modify the
exported output value. All of the imports must be removed before you can delete
the exporting stack or modify the output value.

Consult the [exporting stack output values] guide to review the differences
between importing exported values and using nested stacks. 
```

- Distinguish "outputs" from exported outputs
- Separated outputs to another paragraph and made it clear not all values are available as exports

matteofigus: Makes sense, thanks!

ctd (approved):

PR #239, opened 2021-03-02
Remove queue items limit for individual jobs

### LINKED ISSUE #198: Increase size of the deletion queue for individual jobs
After reviewing #188 we kicked off a discussion to start increasing the size of the deletion queue for individual jobs. I decided to summarise the main points in this issue so we can define a strategy moving forward.

## Problem statement

In order to deal with specific internal limits (DynamoDB doc size, Athena query size, SQS message size, and API payload size), currently there is a limit of number of queue items picked by a deletion job. If the queue is too big, a Job will "skip" the items that can't fit and then they will be available to be included in the next Job. Current size of the queue is about ~400KB, which typically means some thousands of matches.

## Why is it a problem?

If you need to process thousands of queue items and run it into multiple jobs, you may incur in extra overhead:

* **Cost**: Athena queries need to be re-run for every job. Fargate needs to re-process the found results multiple times too. If the same S3 object would be affected by multiple jobs, there would be S3+VPC waste too.
* **Performance**: Time can increase significantly as there is lots of repetition
* **Resiliency**: If the find would need to process the same Object, that would need to be rewritten multiple times. Possibility of errors increases by repeating the action multiple times

## What we propose to change to solve the issue?

The new design uses S3 to store the temporary state to represent the queue of a given job (aka Job manifest) rather than DynamoDB. The current list is a summary of the changes to be made to guarantee the current feature set, including having multiple column identifiers for Data Mappers, support for both Parquet/JSON, and expiration policies for Job Details.

1. On CloudFormation stack create, a Glue Db/Table is created used by the solution to look for `s3://manifests-bucket/manifests/` to be used for Athena joins, with partitioning by **job id, data mapper id**
2. On job start, the queue won't be processed just yet to avoid the API to timeout while scanning the queue table and writing to S3. The DynamoDB queue will be written into a S3 JSON object during the query planning phase in the manifests bucket with encryption=ON. The partitions will be then added to the Glue DB/Table.
3. **Rather than writing the matches to the DynamoDB document**, we will write in the Job Document only the path to the S3 Object in the archive:
    ```json
    {
      ...
      "Manifests": "s3://manifests-bucket/manifests/job-id/data-mapperid/manifest.json"
    }
    ```
    * When querying the `API/jobs/id` endpoint during the job running, the `Manifests` attribute will be shown
    * When the job will complete, the manifests will be persisted and kept available to be queried using Athena.
4. If `JobDetailsRetentionDays` is set, a DynamoDB stream will be setup on Job DynamoDB Document DELETE to delete the manifests from S3 + the partitions from glue.
5. The Job processor API endpoint lambda will be rewritten to remove the item skipping logic and the DDB doc size check logic (we shouldn’t need either anymore).
6. The Query builder lambda should be rewritten so that Athena queries make a join rather than inlining the match_ids and should read the queue from the S3 object rather than Dynamo. That lambda should possibly cache that using `@lru_cache()`
7. The Query results handler should store in SQS only the S3 path of the manifest rather than the full matches
8. The Fargate tasks should read the queue from the S3 manifest rather than Dynamo, caching via `@lru_cache()`
9. We should remove the Item skipping from UI
10. In the job details UI, it won't be possible to see the queue anymore after successful job completion.
11. Update LIMITS doc to remove the skipping stuff and deprecate the skipping related job status/job events from troubleshooting

## Summary of changes

* **Access**: Query Handlers Step function lambdas + Fargate + Athena need to access the S3 manifest
* **API**: We'll show `Manifests` for new jobs, we'll show `DeletionQueueItems` for old jobs.
* **Cost**: We pay for storing the queue in S3, but we don’t store it anymore in DDB. In terms of API calls, all that was accessed by Dynamo in terms of queue now is in S3. Costs, for even high numbers, should be kind of the same for each job. Obviously, as soon as we allow running multiple jobs into one, there are very significant Cost advantages with the new approach
* **Docs**: Update limits, document we have a Glue DB/Table used by solution during job creation

## What's next

I definitely would love some input from @ormahler given this design is based on their Pull Request. If this plan works, I can then start breaking it into multiple Pull Requests, cherry picking some stuff from #188 and add all the other bits accordingly (tests, CloudFormation bits, documentation updates, and so on).

matteofigus:
*Issue #, if available:*

Closes #198, #188

*Description of changes:*

This PR removes the limit on the number of queue items to fit an individual job to improve performance and cost.
To solve the issue, the queue is now assigned to a job asynchronously after job start, and its state is saved as snapshot (aka manifest, saved in JSON format in S3) to be consumed by the various components in the Find and Forget Phase.

Here is a summary of the main limits and how they have been addressed in this changes:
1. timeout during job start API call for a big queue: the queue is now populated asynchronously in the generate queries lambda - 15m is the new limit which should be enough for millions of queue items
2. athena query size: the manifests are part of a JOIN with the target tables, meaning the query is not proportional to the queue size
3. get job API size: the queue items are not part of the response anymore. Once the query generation happens, the urls of the manifests (one for each DataMapper) are part of the response, together with _just_ the queue size.
4. step function state size: the state doesn't contain the queue anymore
5. Dynamo job document size: the queue is not snapshotted into a dynamo document anymore, we save it in S3 instead where we have virtually no limit to its size

Inevitably there are some breaking changes to the API and Web UI functionality which have been documented in the upgrade guide.

*UI Changes*

I took the opportunity of using the new data and event (the QueryPlanningComplete) to improve the UI a bit. At the beginning the Items size is just left unknown until the event writes how many queries are planned to execute + the queue size and the manifests:

Screenshot 2021-03-02 at 18 44 55

Now I also can show a bit better the progress during the Find:

Screenshot 2021-03-02 at 18 45 16

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed


backend/lambdas/tasks/generate_queries.py
```diff
@@ -40,20 +50,92 @@
...
                 "Unsupported data mapper query executor: '{}'".format(query_executor)
             )
 
         batch_sqs_msgs(queue, queries)
+        total_queries += len(queries)
+    write_partitions(manifests_partitions)
+    return {
+        "GeneratedQueries": total_queries,
+        "DeletionQueueSize": len(deletion_items),
+        "Manifests": list(
```
HieronymusLex, line 75: nit: list comprehension probably makes more sense here since you're immediately calling list with the iterator from map

docs/UPGRADE_GUIDE.md
```diff
@@ -1,6 +1,40 @@
 # Upgrade Guide
 
-## Migrating from <=v0.8 to 0.9
+## Migrating from <=v0.23 to v0.24
+
+Prior to v0.24, the Deletion Queue was synchronously processed on Job Creation
+and stored in DynamoDB. As a result, the job API provided the full queue for a
+given job in the `DeletionQueueItems` property and there was a limit of ~375KB
+on the queue size for each individual Job (if the size for a given job would
```
HieronymusLex, line 8: nit: don't think the second statement needs to be in brackets. It's a separate sentence

frontend/src/components/StartDeletionJob.js
```diff
@@ -16,16 +16,16 @@ const StartDeletionJob = ({ className, gateway, goToJobDetails }) => {
...
-        gateway.getQueue(),
         gateway.getDataMappers()
       ]);
 
       setSummary({
-        "Deletion Queue Size": MatchIds.length,
         "Data Mappers Count": DataMappers.length,
-        "Job Details Retention (Days)": Settings.JobDetailsRetentionDays,
+        "Job Details Retention Policy": Settings.JobDetailsRetentionDays
+          ? `Retain for ${Settings.JobDetailsRetentionDay} days`
```
HieronymusLex, line 27: this won't make sense if retention is 1 day?

frontend/src/components/pages/DeletionJob.js
```diff
@@ -149,6 +146,8 @@ const DeletionJob = ({ gateway, jobId }) => {
     setExportState("initial");
   };
 
+  const isLegacy = (job) => !isUndefined(job.DeletionQueueItems);
```
HieronymusLex, line 149: Perhaps we can find a better term than `isLegacy`? I'm thinking what if we deprecate a different field in a future release that has no relation to the presence of the `DeletionQueueItems`

templates/manifests.yaml
```diff
@@ -0,0 +1,83 @@
...
+      BucketEncryption:
+        ServerSideEncryptionConfiguration:
+          - ServerSideEncryptionByDefault:
+              SSEAlgorithm: AES256
+      LifecycleConfiguration:
+        Rules:
+          - Id: ExpireDeletedManifests
+            Prefix: manifests/
+            Status: Enabled
+            NoncurrentVersionExpirationInDays: 1
```
HieronymusLex, line 21: question: would there be any value in retaining the manifest beyond 1 day e.g. the same job data retention setting. I suspect not but worth asking!

HieronymusLex (changes requested):

frontend/src/components/pages/DeletionJob.js
matteofigus: I added a little comment and renamed, better?

templates/manifests.yaml
matteofigus: Not sure I understand the question. Currently, the manifests are deleted (and their partitions) by the stream processor when the DDB job document expires, so I think it has the same data retention as the job data currently. This is only for deleting the thing indefinitely as the bucket has versioning enabled - so the deletion actually just creates a deletion marker. I could potentially change the stream to fetch the version first and do a hard deletion but perhaps that would be a bit more expensive (computationally + S3 calls) so that's why I decided to do it like this.

HieronymusLex: doh, I missed the noncurrent bit

HieronymusLex (dismissed):
lgtm - would be worth @ctd also casting an eye over this

HieronymusLex:
Perhaps it should say "Pending Planning" rather than "Planning Pending"?

docs/UPGRADE_GUIDE.md
```diff
@@ -1,6 +1,41 @@
...
+given job in the `DeletionQueueItems` property and there was a limit of ~375KB
+on the queue size for each individual Job. If the size for a given job would
+have exceeded the allowed space, the `DeletionQueueItemsSkipped` property would
+have been set to `true` and it would have been necessary to run one or more
+deletion jobs, upon completion, to process the whole queue.
+
+Starting from v0.25, the queue is processed asynchronously after job creation
+and is stored in S3 in order to remove the queue limit. As a result:
+
+1. The fields `DeletionQueueItemsSkipped` and `DeletionQueueItems` are both
```
ctd, line 16:
```suggestion
1. The `DeletionQueueItemsSkipped` and `DeletionQueueItems` fields are both
```

frontend/src/components/pages/DeletionJob.js
```diff
@@ -207,45 +211,25 @@ const DeletionJob = ({ gateway, jobId }) => {
...
-                )}`}
-              />
-              <span>
-                {job.DeletionQueueItemsSkipped.toString().toUpperCase()}
-              </span>
+            <DetailsBox label="Deletion Queue Size" fullWidth>
+              {isUndefined(job.DeletionQueueSize)
+                ? syncQueueProcessing(job)
+                  ? job.DeletionQueueItems.length
+                  : "Unknown (Planning pending)"
```
ctd, line 218:
```suggestion
                  : "Unknown (Plan pending)"
```

Re: https://github.com/awslabs/amazon-s3-find-and-forget/pull/239#issuecomment-789836706

If the queue size is determined from the product of planning, then that'd make it a plan that we're awaiting on.

ctd (commented):
Sorry for the late review. This looks really good, there's just a few small suggestions from me.

backend/lambda_layers/boto_utils/python/boto_utils.py
```diff
@@ -236,3 +237,27 @@ def get_session(assume_role_arn=None, role_session_name="s3f2"):
...
+
+def fetch_job_manifest(path):
+    bucket, obj = parse_s3_url(path)
+    return s3.Object(bucket, obj).get().get("Body").read().decode("utf-8")
+
+
+def json_lines_iterator(content, include_unparsed=False):
+    lines = content.split("\n")
+    if lines[-1] == "":
+        lines.pop()
```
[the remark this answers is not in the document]
matteofigus: Aha, that is, I think, what we used to do (this piece of code is here being moved to boto utils from the ecs tasks part). I remember a bug related to having new lines as part of the content, and managed to reproduce in a unit test: https://github.com/awslabs/amazon-s3-find-and-forget/blob/master/tests/unit/ecs_tasks/test_json.py#L71-L88 

In conclusion, I think the `splitlines` method is the issue, which I fixed in this way. I retried to apply this suggested change, and in fact that test indeed fails.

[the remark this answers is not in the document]
ctd: ok, i wasn't aware -- fair enough!

ctd (dismissed):

ctd (approved):

HieronymusLex (approved):
LGTM - great work

ormahler:
This is amazing!
@matteofigus Great work.
Would appreciate if you add me as a contributor to the project if you find it appropriate :).

matteofigus:
Thanks @ormahler - your contribution was key for sure - thanks a lot for your help! Sure I find it appropriate, I'll see what I can do, I guess I can add a note somewhere.

ormahler:
Sounds Good, Thanks!

PR #240, opened 2021-03-03
Improvements on Fargate service

matteofigus:
*Description of changes:*

* Add ECR API to support Fargate Platform 1.4.0 (we currently use `LATEST` which currently points to 1.3.0 and will be resolving 1.4.0 soon)
* Remove custom resource for setting up container insights as now it's possible to declare the setting natively via CloudFormation

I tested the code by setting the platform version to 1.4.0 manually and making sure all worked for unit and acceptance. ~~I then left the platform version unchanged (so pointing to LATEST) as I think that is best for the future - but perhaps we can pin it to 1.4.0? Not sure here.~~ I then decided to pin to 1.4.0 to make migration explicit. We can move back to `LATEST` later on when 1.4.0 is fully rolled out.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed


docs/USER_GUIDE.md
```diff
@@ -42,12 +42,15 @@ The Fargate tasks used by this solution to perform deletions must be able to
 access the following AWS services, either via an Internet Gateway or via [VPC
 Endpoints]:
 
-- Amazon S3
-- Amazon DynamoDB
-- Amazon CloudWatch (monitoring and logs)
-- AWS ECR
-- Amazon SQS
-- AWS STS
+- Amazon S3 gateway endpoint (_com.amazonaws.**region**.s3_)
```
HieronymusLex, line 45: the above sentence no longer makes sense with this updated list
> The Fargate tasks used by this solution to perform deletions must be able to
access the following AWS services, either via an Internet Gateway or via [VPC
Endpoints]:

The new list is a list of endpoints rather than services

HieronymusLex (changes requested):
I'm fine with this change but I'm not 100% clear why are splitting the work into 2 steps by pinning to a specific version then later pinning to LATEST? Is this problem that under 1.3.0, this change will not work? So until LATEST resolves to 1.4.0, the solution provided VPC would be unusable? Or something else? For example, what is the advantage of pinning this now vs delaying this change till 1.4.0 becomes latest?

docs/USER_GUIDE.md
matteofigus: You're right - updated, better?

matteofigus:
> I'm fine with this change but I'm not 100% clear why are splitting the work into 2 steps by pinning to a specific version then later pinning to LATEST? Is this problem that under 1.3.0, this change will not work? So until LATEST resolves to 1.4.0, the solution provided VPC would be unusable? Or something else? For example, what is the advantage of pinning this now vs delaying this change till 1.4.0 becomes latest?

So summarising, 1.3.0 works right now but 1.4.0 needs the new VPC endpoint. When LATEST will resolve to 1.4.0 if you won't have this version of the solution, a new deploy will fail just fail.
My initial thinking was to just add the endpoint and point customers to the newest version of the solution. Existing deployments, would probably be fine as LATEST would have resolved 1.3.0 at deploy time and that wouldn't change over time.

Then I tried to think that perhaps there are people that want to test if this works before AWS just changes the resolving mechanism (perhaps in a non-prod environment) - so then I thought that pinning to next just makes that easier. In regards of keeping the version pinned or switching back to `LATEST`, I am not sure, that's why I was suggesting to leave that to the future.

HieronymusLex (approved):
lgtm

PR #245, opened 2021-04-28
Support for CSE-KMS

matteofigus:
*Description of changes:*
This PR is for adding support for CSE-KMS.
To summarise, the idea is to handle objects that are CSE with KMS by decrypting them and re-encrypting them back with the same configuration (same key, same algorithm).

A couple of design choices:
1. The only way supported for CSE-KMS is the same used by Athena, consisting on using Objects' headers to store the encryption envelope, compatible with what supported official SDKs do today
2. I added support only for V2 encrypted objects (they have a `x-amz-key-v2` header rather than `x-amz-key` header)
3. When encrypting back, I use the same strategy found when decrypting, so same KMS key, and same algorithm (GCM or CBC)

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed


backend/ecs_tasks/delete_files/cse.py
```diff
@@ -0,0 +1,138 @@
...
+
+def is_kms_cse_encrypted(s3_metadata):
+    if HEADER_KEY in s3_metadata:
+        if s3_metadata.get(HEADER_WRAP_ALG, None) != "kms":
+            raise ValueError("Unsupported Encryption strategy")
+        if s3_metadata.get(HEADER_ALG, None) not in [ALG_CBC, ALG_GCM]:
+            raise ValueError("Unsupported Encryption algorithm")
+        return True
+    elif "x-amz-key" in s3_metadata:
+        raise ValueError("Unsupported Encryption SDK version")
```
HieronymusLex, line 35: which SDK is this referring to?

templates/deletion_flow.yaml
```diff
@@ -187,6 +187,11 @@ Resources:
           - Action: s3:GetObject*
             Effect: Allow
             Resource: !Sub arn:aws:s3:::${ManifestsBucket}/manifests/*
+          - Action:
+            - kms:Decrypt
+            - kms:GenerateDataKey*
+            Effect: Allow
+            Resource: !Sub arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/*
```
HieronymusLex, line 194: can we scope this down at all? what about keys which aren't in the current account? Feels like we might need to parametrise this somehow

HieronymusLex (changes requested):

ctd (commented):
Looks good to me, though might be a good idea to have the encryption methods reviewed by someone with some crypto expertise?

backend/ecs_tasks/delete_files/cse.py
matteofigus: Ah good point, basically the SDKs used to encrypt in a way called v1 and then switched to v2: https://github.com/aws/aws-sdk-java/tree/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/crypto

The v2 is better as it supports CBC (aka unauthenticated encryption) and GCM (aka authenticated encryption)  while v1 was CBC only. You can recognise as one has the `x-amz-key` header and one has `x-amz-key-v2`.

In conclusion, not sure how I could name it, Encryption SDK could be confused with [this](https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/introduction.html) which is actually unrelated. 

In [this snippet](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingClientSideEncryption.html) they call the variables "AmazonS3EncryptionClientV2".

The best way I can think of could be `Unsupported S3 Encryption Client Version`?
```suggestion
        raise ValueError("Unsupported S3 Encryption Client Version")
```

templates/deletion_flow.yaml
matteofigus: Good point about cross-account! 
It could be messy unless I parametrise the arns in CFN I guess.

We have to consider that you still need to add the solution task arn in each key's policy as suggested by the policy generator, so I wonder if that is ok? I don't like wildcards but still, you need to go key by key anyways, would that be enough?

matteofigus:
After more code reviews, I intend to make this changes:
1) Restrict IAM action `kms:GenerateDataKey*` to `kms:GenerateDataKey` as that's enough for symmetric keys (which are the only ones supported by Athena)
2) Spike creating a KMS key for testing that gets reused for acceptance, rather than creating and scheduling for deletion one for each relevant test
3) Spike using moto rather than the custom implementation I used for mocking KMS during unit

templates/deletion_flow.yaml
matteofigus: Ok so I decided to parametrise the Key Arns, that allowed me to remove all wildcards in the policies as well as solving the cross-account issue.

ctd (approved):

HieronymusLex (dismissed):

HieronymusLex (dismissed):

ctd (approved):
👏

PR #257, opened 2021-07-22
Allow to specify granularity for partition querying

matteofigus:
*Description of changes:*

This is a PR to enable customers to customise the query strategy by declaring the Partition Keys to be used for generating queries.

Let's take an example a large dataset with Partition Keys=[year,month,day] as 3 individual partition keys.
The solution as today will fetch the cartesian product of them and will perform a query for each unique combination of values. 

For example, let's assume daily data starting Feb 1st 2020 and ending March 1st 2021.
With this new feature, a Data Mapper will be able to specify none or any of the Partition Keys:
* With `PartitionKeys=[]` there will be 1 query (no partition keys used)
* With `PartitionKeys=["year"]` there will be 2 queries (y=2020 + y=2021)
* With `PartitionKeys=["month"]` there will be 12 queries
* With `PartitionKeys=["year","month"]` there will be 13 queries
* etc..
* By omitting `PartitionKeys` we'll have a default of all, equivalent to `["year","month","day"]` for backward compatibility (365+28 if I'm doing the math right 😂 )

This feature was requested by customers with very large amount of very small partitions - that currently need to wait hours for a deletion job to complete, that given the very small amount of data should be able to query without partitioning in minutes instead - therefore saving in athena concurrency and overall lambda cost.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed

### DIFF
--- a/docs/USER_GUIDE.md
+++ b/docs/USER_GUIDE.md
@@ -398,8 +398,23 @@ To grant these permissions in Lake Formation:
    have a greater number of smaller queries (the same query will be repeated
    with a `WHERE` additional clause for each combination of partition values).
    If you have a lot of small partitions, it may be more efficient to choose
-   none or a subset of partition keys from the list. If instead you have very
-   big partitions you may want to select all the partition keys.
+   none or a subset of partition keys from the list in order to increase speed
+   of execution. If instead you have very big partitions, it may be more
+   efficient to choose all the partition keys in order to reduce probability of
+   failure caused by query timeout. As a rule of thumb, we recommend the average
+   query size to be between 100GB and 1TB.
+
+   > As an example, let's consider 10 years of daily data with partition keys of
+   > `year`, `month` and `day` with total size of `100TB`. By declaring
+   > PartitionKeys=`[]` (none) a single query of 100TB would run during the Find
+   > phase, and that may be too much to complete within the 30m limit of Athena
+   > execution time. On the other hand, using all the combinations of the
+   > partition keys we would have approximately `3650` queries, each being
+   > probably very small, and given the default Athena concurrency limit of
+   > `20`, it may take very long to execute all of them. The best in this
+   > scenario is possibly the `['year','month']` combination, which would result
+   > in `120` queries.
+
 6. From the columns list, choose the column(s) the solution should use to to
    find items in the data which should be deleted. For example, if your table
    has three columns named **customer_id**, **description** and **created_at**
--- a/docs/USER_GUIDE.md
+++ b/docs/USER_GUIDE.md
@@ -401,8 +401,8 @@ To grant these permissions in Lake Formation:
    none or a subset of partition keys from the list in order to increase speed
    of execution. If instead you have very big partitions, it may be more
    efficient to choose all the partition keys in order to reduce probability of
-   failure caused by query timeout. As a rule of thumb, we recommend the average
-   query size to be between 100GB and 1TB.
+   failure caused by query timeout. We recommend the average query size to be
+   between 100GB and 1TB.
 
    > As an example, let's consider 10 years of daily data with partition keys of
    > `year`, `month` and `day` with total size of `100TB`. By declaring
--- a/docs/USER_GUIDE.md
+++ b/docs/USER_GUIDE.md
@@ -401,15 +401,15 @@ To grant these permissions in Lake Formation:
    none or a subset of partition keys from the list in order to increase speed
    of execution. If instead you have very big partitions, it may be more
    efficient to choose all the partition keys in order to reduce probability of
-   failure caused by query timeout. We recommend the average query size to be
-   between 100GB and 1TB.
+   failure caused by query timeout. We recommend the average query size not to
+   exceed the hundreds of GBs and not to take more than 5 minutes.
 
    > As an example, let's consider 10 years of daily data with partition keys of
-   > `year`, `month` and `day` with total size of `100TB`. By declaring
-   > PartitionKeys=`[]` (none) a single query of 100TB would run during the Find
-   > phase, and that may be too much to complete within the 30m limit of Athena
-   > execution time. On the other hand, using all the combinations of the
-   > partition keys we would have approximately `3650` queries, each being
+   > `year`, `month` and `day` with total size of `10TB`. By declaring
+   > PartitionKeys=`[]` (none) a single query of `10TB` would run during the
+   > Find phase, and that may be too much to complete within the 30m limit of
+   > Athena execution time. On the other hand, using all the combinations of the
+   > partition keys we would have approximately `3652` queries, each being
    > probably very small, and given the default Athena concurrency limit of
    > `20`, it may take very long to execute all of them. The best in this
    > scenario is possibly the `['year','month']` combination, which would result
--- a/docs/USER_GUIDE.md
+++ b/docs/USER_GUIDE.md
@@ -392,24 +392,32 @@ To grant these permissions in Lake Formation:
 4. Select a **Query Executor Type** then choose the **Database** and **Table**
    in your data catalog which describes the target data in S3. A list of columns
    will be displayed for the chosen Table.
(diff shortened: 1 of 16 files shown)


docs/USER_GUIDE.md
```diff
@@ -392,24 +392,32 @@ To grant these permissions in Lake Formation:
...
-5. From the list, choose the column(s) the solution should use to to find items
-   in the data which should be deleted. For example, if your table has three
-   columns named **customer_id**, **description** and **created_at** and you
-   want to search for items using the **customer_id**, you should choose only
-   the **customer_id** column from this list.
-6. Enter the ARN of the role for Fargate to assume when modifying objects in S3
+5. From the Partition Keys list, select the partition key(s) that you want the
+   solution to use when generating the queries. If you select none, only one
+   query will be performed for the data mapper. If you select any or all, you'll
+   have a bigger number of smaller queries (the same query will be repeated with
```
HieronymusLex, line 398:
```suggestion
   have a greater number of smaller queries (the same query will be repeated with
```

docs/USER_GUIDE.md
```diff
@@ -392,24 +392,32 @@ To grant these permissions in Lake Formation:
...
+   a `WHERE` additional clause for each combination of partition values). If you
+   have a lot of small partitions, you may choose none or few partition keys
```
HieronymusLex, line 400:
```suggestion
   have a lot of small partitions, it may be more efficient to choose none or a subset of partition keys
```

HieronymusLex: we should probably say why this is the case too

docs/USER_GUIDE.md
```diff
@@ -392,24 +392,32 @@ To grant these permissions in Lake Formation:
...
+   from the list. If instead you have very big partitions you may want to select
```
HieronymusLex, line 401: same as above, probably want to say it may be more efficient

frontend/src/components/pages/DataMappers.js
```diff
@@ -191,38 +191,43 @@ const DataMappers = ({ gateway, onPageChange }) => {
...
+                return (
+                  <tr
+                    key={index}
+                    className={selectedRow === index ? "selected" : undefined}
+                  >
+                    <td style={{ textAlign: "center" }}>
+                      <Form.Check
+                        inline
+                        type="radio"
+                        id={`inline-${index}`}
```
HieronymusLex, line 206: nit: the ID should probably be indicative of what the field is e.g. dm-${index}

frontend/src/components/pages/NewDataMapper.js
```diff
@@ -132,20 +166,23 @@ const NewDataMapper = ({ gateway, goToDataMappers }) => {
...
+    let checkboxes = document.getElementsByName("column");
+    for (let i = 0; i < checkboxes.length; i++) checkboxes[i].checked = false;
 
-  const columnsForSelectedTable = selectedTable ? selectedTable.columns : [];
-  const noTables = !glueData || isEmpty(glueData.databases);
+    const partitionKeys = selectedTable?.partitionKeys || [];
+    setPartitionKeys({ type: "reset", value: partitionKeys });
+    checkboxes = document.getElementsByName("partition-key");
+    for (let i = 0; i < checkboxes.length; i++)
+      checkboxes[i].checked = !isEmpty(partitionKeys);
```
HieronymusLex, line 184: what's happening here? seems not v react-y to be setting the element prop like this rather than via state?

frontend/src/components/pages/NewDataMapper.js
```diff
@@ -326,21 +356,36 @@ const NewDataMapper = ({ gateway, goToDataMappers }) => {
...
               </Form.Group>
+              <Form.Group>
+                <Form.Label>Partition Keys used to generate queries</Form.Label>
+                <Form.Text className="text-muted">
+                  {isUndefined(selectedTable)
+                    ? `No table selected`
+                    : isEmpty(partitionKeysForSelectedTable)
+                    ? `None - the table is not partitioned`
+                    : `In order to control the granularity of each query, you can
+                      select none or more partition keys to be used. If you
```
HieronymusLex, line 367:
```suggestion
                      select the partition keys to be used in the query phase. If you
```

frontend/src/components/pages/NewDataMapper.js
```diff
@@ -326,21 +356,36 @@ const NewDataMapper = ({ gateway, goToDataMappers }) => {
...
+                  partitionKeys={partitionKeysForSelectedTable}
+                  setPartitionKeys={setPartitionKeys}
+                />
+              </Form.Group>
               <Form.Group>
                 <Form.Label>Columns used to query for matches</Form.Label>
                 <Form.Text className="text-muted">
-                  Select one or more column from the table
+                  {!isEmpty(columnsForSelectedTable)
+                    ? `Select one or more column from the table`
```
HieronymusLex, line 381:
```suggestion
                    ? `Select one or more columns from the table`
```

HieronymusLex (changes requested):
nice work. some minor comments

docs/USER_GUIDE.md
matteofigus: Done, I added an example too.

frontend/src/components/pages/NewDataMapper.js
matteofigus: So the issue is that when a state consists of an array or object in general, react doesn't work great out of the box because you cannot control mutations as you would with a flattened state.
Let's take the example of:
```
const [foo, setFoo] = useState([]);
foo.push("c");
setFoo(foo);
```
In this case, because `oldFoo === newFoo` the `setFoo` won't trigger a re-render.

So in this particular case, I am basically:
1) using a reducer to set the initial state value (with the reset) and making sure the frontend reflects that (checking/unchecking the checkboxes). 
2) Once you make changes in the UI (you check/uncheck the boxes) that updates the state (so, basically unidirectional)
3) If you unselect a table or you select a different one, I recall the reset because I want all the partition keys to be selected as default

HieronymusLex: you shouldn't mutate foo directly though? your example should be:
```
const [foo, setFoo] = useState([]);
setFoo(foo.concat(["c"]);
```

Then checked state on the checkbox should be set by `foo.indexOf("c") !== -1`. onChange handler for the checkbox would be something like: 
```
<Checkbox onChange={toggleCheckbox("c")} ..../>

const toggleCheckbox = val => e => {
  const isChecked = event.target.checked;
  setFoo(isChecked
    ? foo.filter(i => i !== val))
    : foo.concat([val])
  )
}
```

ctd (commented):
looks good, just two small copy change suggestions from me

frontend/src/components/pages/NewDataMapper.js
matteofigus: Ok so I think this is clear, the change handler you suggest is `useState` based and I am using a reducer, and I think I prefer the reducer but that doesn't make such difference.

I think the ugliness consists on the fact the checkboxes didn't have a `checked` property and I was setting the initial state by checking the ID and selecting all via JS for Partition Keys and empty for columns; but I changed that now.
I am not sure why I did this when working on the columns, but I think I found a cleaner way now if you look at the new diff:
1. I have a reducer that I can use for any multi-value based array (which I have in fact moved to utils)
2. For both the `ColumnsViewer` and `PartitionKeysViewer` I am actually now using a `checked` property to get the initial state from state now
3. I have a `reset` function that allows me to set the defaults when I change the selected table (all for PKs, none for columns)
4. As you can see from the diff, I was before resetting the defaults in the `onChange` handler for the db/table dropdowns, but now I moved it to a `useEffect` as I think that's cleaner and works well.

Let me know what do you think of it now, I think it's much better and full react-y :D now

ctd (approved):

HieronymusLex (approved):

PR #274, opened 2021-11-04
Fix parquet multiple column identifier bug

guvenbz:
*Issue #, if available:*

*Description of changes:*

Deletion from some parquet files fails on [this line](https://github.com/awslabs/amazon-s3-find-and-forget/blob/master/backend/ecs_tasks/delete_files/parquet_handler.py#L131) with the following error:

> TypeError: ufunc 'invert' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

The issue appears to be that `table` gets emptied out on the first iteration, so on the second iteration `indexes` is empty.

Verified that this PR fixes the issue.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## v0.37 (unreleased)
+
+- [#274](https://github.com/awslabs/amazon-s3-find-and-forget/pull/274):
+  Fix for a bug that causes deletion to fail in parquet files when a data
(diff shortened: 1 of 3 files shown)


matteofigus (approved):
Great, LGTM, thanks for the PR

PR #276, opened 2021-11-05
Auto-refresh access tokens used by containers

matteofigus:
*Description of changes:*

This is a PR to fix a bug observed when processing takes more than an hour, causing the expiration of an access token. This is an example of a exception:

```
2021-11-02T07:17:27.424-07:00	[INFO] Generating new file without matches
2021-11-02T08:18:06.356-07:00	[ERROR] ClientError: An error occurred (ExpiredToken) when calling the GetBucketRequestPayment operation: The provided token has expired.
```

I investigated increasing the token validity to 12 hours, but unfortunately the maximum duration for temporary credentials is 1 hour. I decided to setup a mechanism to auto-refresh tokens but then I bumped into this [blogpost](https://ben11kehoe.medium.com/boto3-sessions-and-why-you-should-use-them-9b094eb5ca8e), which also came with a handy python library that I manually tested and worked perfectly (I manually created a 15m sleep with a token validity of 15m and it correctly performed the refresh and succeeded).

*PR Checklist:*

- [x] Changelog updated
- [x] Unit tests (and integration tests if applicable) provided
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #278, opened 2021-11-10
Make sure s3fs uses the session rather than stale token after expiration

matteofigus:
*Issue #, if available:*
In #276 I managed to solve the expiration token issue happening when using the s3 boto client by using the library that wraps the session handling. Unfortunately, when doing more testing, I bumped into a new bug: `ObjectUpdateFailed` with `"Unable to retrieve object: Write failed: TypeError(\"'NoneType' object is not subscriptable\")"`.

After doing some digging, I found that while the previous PR solved the issue by refreshing the token from the client (https://github.com/awslabs/amazon-s3-find-and-forget/blob/master/backend/ecs_tasks/delete_files/s3.py#L18-L21), the S3fs client was still using stale credentials. By passing in the session to the s3fs factory, I verified the issue is fixed by artificially creating a 1h sleep and waiting for it to be reproduced and then fixed.
Sorry for not catching this earlier. In my previous test I only tested for 15m expiration, which didn't catch this scenario.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@
 ## v0.38
 
 - [#278](https://github.com/awslabs/amazon-s3-find-and-forget/pull/278): Fix for
-  a bug that causes the access token to expire and cause a Job to fail if
-  processing of an object takes more than an hour
+  a bug that caused a job to fail if the processing of an object took longer
+  than the lifetime of its IAM temporary access credentials.
 
 ## v0.37
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,16 @@
 # Change Log
 
-## v0.37
+## v0.38
 
-- [#276](https://github.com/awslabs/amazon-s3-find-and-forget/pull/276): Fix for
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,11 +1,17 @@
 # Change Log
 
-## v0.37
+## v0.38
 
-- [#276](https://github.com/awslabs/amazon-s3-find-and-forget/pull/276): Fix for
+- [#278](https://github.com/awslabs/amazon-s3-find-and-forget/pull/278): Fix for
   a bug that causes the access token to expire and cause a Job to fail if
   processing of an object takes more than an hour
```
ctd, line 7:
```suggestion
  a bug that caused a job to fail if the processing of an object took longer
  than the lifetime of its IAM temporary access credentials.
```

ctd (commented):
Small nitpick about the changelog description, otherwise LGTM.

PR #279, opened 2021-11-16
Improve performance for large queues of simple matches

matteofigus:
*Description of changes:*

When processing a job with a large queue, there may be some performance issues in the Forget phase. In particular, with a queue of 10K matches in a dataset of 2 objects of 300MB each, I was able to complete the job only after 50m, as opposite of a couple of minutes when the same dataset had 1 match in the queue (of which the actual compute was a bunch of seconds).

After this improvement, I was able to obtain the same result for the large queue. Note that this helps only with simple matches, as composite still iterate over lists. I plan to address that in a separate PR as I think that is a bit more complex and I need more time to think about it.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,15 @@
 # Change Log
 
+## v0.39
+
+- [#279](https://github.com/awslabs/amazon-s3-find-and-forget/pull/279): Improve
+  performance for large queues of simple matches
+
 ## v0.38
 
 - [#278](https://github.com/awslabs/amazon-s3-find-and-forget/pull/278): Fix for
   a bug that caused a job to fail if the processing of an object took longer
-  than the lifetime of its IAM temporary access credentials.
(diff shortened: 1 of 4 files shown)


ctd (commented):
I think it could be better to perform the cast to a set inside the get_row_indexes_to_delete function, just on the principal to not lose this optimisation if the function were to be invoked elsewhere.

Also the changelog should mention the logging additions.

matteofigus:
> I think it could be better to perform the cast to a set inside the get_row_indexes_to_delete function, just on the principal to not lose this optimisation if the function were to be invoked elsewhere.
> 
> Also the changelog should mention the logging additions.

Good idea, done.

ctd (approved):
LGTM

PR #287, opened 2021-11-30
Add parameter for ignoring Object Not Found exceptions during deletion

guvenbz:
*Description of changes:*
Currently a deletion job will fail if objects that were found in the Find phase no longer exist in the Delete phase. This can cause spurious failures when lifecycle policies are configured on a bucket which cause objects to expire.

This PR adds a data mapper parameter to allow users to ignore these errors. The default behavior remains the same.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## v0.43 (unreleased)
+
+- [#287](https://github.com/awslabs/amazon-s3-find-and-forget/pull/287): Add
+  data mapper parameter for ignoring Object Not Found exceptions encountered
+  during deletion
+
 ## v0.42
 
 - [#285](https://github.com/awslabs/amazon-s3-find-and-forget/pull/285): Fix for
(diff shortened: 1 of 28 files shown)


ctd:
It looks like this new functionality as-is would skip deleting old S3 object versions when an object is skipped.

We should either correct this or make it very clear in the documentation so that users don't keep data they intended to delete due to skipping.

guvenbz:
Great point. My inclination would be to go with the documentation approach, mainly because any systems that are deleting objects from the bucket must handle deletion of old versions anyway.

The user can’t rely on S3F2 to handle this case in general since S3F2 only searches the latest version of an object. Any objects that were deleted before the job began, or before that particular find query ran, etc. would have the same issue. It's just a result of "unfortunate" timing of the deletion that some of these objects end up in the deletion queue.

Does that seem reasonable? I’m happy to do either one if you think otherwise.

guvenbz:
Rebased, updated changelog to reflect newly-released version, added documentation to make it clear that old versions will not be deleted.

ctd (approved):
LGTM. Thanks for your contribution!

PR #293, opened 2022-02-17
Upgrade dependencies

matteofigus:
*PR Checklist:*

- [ ] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass


.github/workflows/publish.yaml
```diff
@@ -45,7 +45,7 @@ jobs:
       - name: Set up Nodejs 12
         uses: actions/setup-node@v1
         with:
-          node-version: 12
+          node-version: 14
```
matteofigus, line 48: Create React App v5 includes loads of warnings for node 12 so I moved all to node 14.

Makefile
```diff
@@ -50,13 +50,13 @@ format-cfn:
 
 .PHONY: format-docs
 format-docs:
-	npx prettier-eslint ./*.md ./docs/*.md --write --prose-wrap always
+	npx prettier ./*.md ./docs/*.md --write
```
matteofigus, line 53: Create React App v5 moved internally to ESLint 8, and prettier-eslint cli is stuck to ESLint 5: https://github.com/prettier/prettier-eslint-cli/issues/304

When upgrading, prettier-eslint was failing. I decided to just move to prettier cli, which has a `--write` functionality as well. The difference is that instead of inferring the linting config from the Create React App module, I am defining the config inside the root's `package.json`. But the advantage is that the dependency tree is much smaller and that allows to remove some vulnerabilities too.

When running with just prettier, the linting is 99% same. You'll notice a couple of new line changes in a couple of files, but I decided that was ok for now.

tests/acceptance/conftest.py
```diff
@@ -140,7 +140,7 @@ def glue_columns():
     ]
 
 
-@pytest.fixture(scope="session", autouse=True)
+@pytest.fixture(scope="module", autouse=True)
```
matteofigus, line 143: I had an issue here when running the acceptance tests. They ran in 64 minutes, and everything running on the last 4 failed because the creds were expired. So I decided to move the cognito token fetch from session to module in order to reduce the probability of that happening again.

ctd (approved):

PR #303, opened 2022-04-04
Improve performance for query generation task

matteofigus:
*Description of changes:*

When generating the queries, the current logic iterates over the data mappers and casts each match id by calling `cast_to_type`, which itself calls `get_column_info` to fetch info on the column to get the type, which itself always calls `column_mapper` to do stuff that is computationally heavy such as creating a tree that defines the schema, which is important for complex data types (types such as `struct`, `array`, `map`).

With this new change, the columns of each data mapper are converted to a tree calling `column_mapper` only once for each data mapper, so that `get_column_info` can lookup and traverse the tree without repeating the operation for each match id.

The result of this optimisation is important when processing hundreds or thousands of matches.
Before, `column_mapper` was called `len(data mappers)*len(deletion queue)`, now just `len(data mappers)`.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## v0.45 (unreleased)
 
+- [#303](https://github.com/awslabs/amazon-s3-find-and-forget/pull/303): Improve
+  performance of query generation step
 - [#301](https://github.com/awslabs/amazon-s3-find-and-forget/pull/301): Include
   table name to error when query generation fails due to an invalid column type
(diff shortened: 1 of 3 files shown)


matteofigus:
@HieronymusLex as discussed, I made some changes to avoid making modifications to the `table` object

HieronymusLex (dismissed):

PR #306, opened 2022-04-12
Add retries for S3 bulk object deletion calls

ctd:
*Issue #, if available:* n/a

*Description of changes:*

boto3 does not appear to retry when individual objects encounter
transient errors during a bulk deletion. This can lead to some jobs
ending in a partial failure due to the old objects returning an error.

This adds a simple retry mechanism where the delete_objects call will be
retried whenever there are any errors returned, with an exponential
backoff, with a maximum number of 10 attempts. Because the API will
return success for deleting an object that is already deleted, there is
no danger to re-attempting deletions that have already succeeded.

Known deficiencies:

This implementation doesn't differentiate the type of error returned.
The pitfall of this is that we might spend extra time retrying errors
that will always be errors (e.g. bad permissions). We could avoid this
by checking only for the presence of particular error codes according
to the [documented error codes](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Error.html)
but I'm not sure it's really worth the added complexity.

The retry parameters are hardcoded. I don't know if it's worthwhile
making these configurable (and where we would put the configuration).

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed


backend/ecs_tasks/delete_files/requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile
+# This file is autogenerated by pip-compile with python 3.8
```
matteofigus, line 2: Do we need Python 3.8 here? If we are moving to Python 3.8, can you upgrade the Local Development Guide, the Github Actions and the Docker image too?

tests/unit/ecs_tasks/test_s3.py
```diff
@@ -606,7 +606,49 @@ def test_it_raises_for_deletion_errors(paginate_mock):
...
+            "Errors": [
+                {"VersionId": "v1", "Key": "key", "Message": "InternalServerError"}
+            ]
+        },
+        {"Errors": [],},
+    ]
+
+    delete_old_versions(s3_mock, "bucket", "key", "v4")
+
+    s3_mock.delete_objects.call_count == 2
```
matteofigus, line 620:
```suggestion
    assert s3_mock.delete_objects.call_count == 2
```

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## v0.45
```
matteofigus, line 3:
```suggestion
## v0.46
```

backend/ecs_tasks/delete_files/s3.py
```diff
@@ -237,6 +233,18 @@ def delete_old_versions(client, input_bucket, input_key, new_version):
         raise DeleteOldVersionsError(errors=[str(e)])
 
 
+@retry(
+    wait=wait_exponential(multiplier=1, min=1, max=10),
+    stop=stop_after_attempt(10),
+    retry=(retry_if_result(lambda r: len(r.get("Errors", [])) > 0)),
+    retry_error_callback=lambda r: r.outcome.result(),
```
matteofigus, line 240: What's your thoughts on adding a log statement in case of retry? I think it would be nice to facilitate troubleshooting especially for the situations where the retries won't fix: https://github.com/jd/tenacity#before-and-after-retry-and-logging

matteofigus (changes requested):

backend/ecs_tasks/delete_files/requirements.txt
ctd: I noticed this too but just disregarded it as a change that must've happened in pip-compile, and I suppose I was running it under python 3.8.

backend/ecs_tasks/delete_files/s3.py
ctd: I'm all for it. I'll add an after= log.

templates/template.yaml
```diff
@@ -1,6 +1,6 @@
 AWSTemplateFormatVersion: "2010-09-09"
 Transform: AWS::Serverless-2016-10-31
-Description: Amazon S3 Find and Forget (uksb-1q2j8beb0) (version:v0.45)
+Description: Amazon S3 Find and Forget (uksb-1q2j8beb0) (version:)
```
matteofigus, line 3: there must be something interesting happening here as this should stay 0.45 if you are not bumping the version. Are your pre-commit scripts running correctly?

backend/ecs_tasks/delete_files/requirements.txt
matteofigus: Ok let's ignore it for now then.

matteofigus (changes requested):

templates/template.yaml
ctd: they aren't. 😀

matteofigus (approved):

PR #308, opened 2022-04-13
Use workspaces and upgrade dependencies

matteofigus:
New NPM allows `overrides` to override specific dependencies. Unfortunately it works only in a root `package.json` when working with monorepos. In order for NPM to use the `overrides`, I am basically converting the root's package to act as monorepo's manifest including the reference to the `frontend` code as workspace. This means that we only have one `package-lock` as NPM does its magic to connect, dedupe and simlink things.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,12 +45,12 @@
       }
     },
     "node_modules/@aws-amplify/analytics": {
-      "version": "5.2.4",
-      "resolved": "https://registry.npmjs.org/@aws-amplify/analytics/-/analytics-5.2.4.tgz",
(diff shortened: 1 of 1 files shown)


ctd:
One question, really - how new does npm need to be to use this? Have you tested it works fine in the deployment codebuild jobs?

matteofigus:
> One question - how new does npm need to be to use this? Have you tested it works fine in the deployment codebuild jobs?

Node 14 ships with npm 8 which supports this. All the stuff depending on npm honestly is done in the github actions. The codebuild job only uses the AWS CLI to get the zip file uploaded to S3 by github actions and unzip it and copy it to the customer bucket. The same command (build) is ran on this PR to run frontend unit tests and it worked (on the last PR, where I fixed the scripts as you can see from the last commit).

matteofigus:
> Let me try fixing this.

It worked now. Good catch 😅

ctd (approved):

PR #310, opened 2022-04-19
Don't fetch column schema during get_partitions call on query generation perfomance

### LINKED ISSUE #309: GenerateQueries lambda times out on highly partitioned tables with many columns
For highly* partitioned tables with many* columns, the GenerateQueries lambda consumes most of its time waiting for [glue get_partitions() calls](https://github.com/awslabs/amazon-s3-find-and-forget/blob/master/backend/lambdas/tasks/generate_queries.py#L268-L274). Scaling the number of partitions in a dataset with many columns causes the lambda to hit the timeout cap of 15 minutes.

On these kinds of tables, get_partitions() returns a very large response payload, a lot of which is unnecessary column metadata. 

One quick fix would be to call get_partitions() with the `ExcludeColumnSchema=True` parameter to reduce the response payload size.

\* For reference, we encountered timeout issues on a table with 800k+ partitions and 170 columns. Applying this change in our specific case sped up execution by ~4x.

matteofigus:
*Issue #, if available:*
Closes #309 

In order to make the updated call using the AWS SDK, I had to update the aws client, which forced me to upgrade other dependencies, including `black` that decided to re-format some python code 😢  
The actual change here is a minimal extra param in the generate queries python task.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed


ctd (approved):
LGTM

PR #313, opened 2022-05-23
IAM Auth and Optional WebUI

danjhd:
*Issue #, if available:* not applicable

*Description of changes:* 
Made the WebUI an optional component.
Adding new option to select IAM for authentication method, which is only valid if WebUI is not deployed.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed


Makefile
```diff
@@ -116,7 +118,7 @@ redeploy-containers:
 redeploy-frontend:
 	$(eval WEBUI_BUCKET := $(shell aws cloudformation describe-stacks --stack-name S3F2 --query 'Stacks[0].Outputs[?OutputKey==`WebUIBucket`].OutputValue' --output text))
 	make build-frontend
-	cd frontend/build && aws s3 cp --recursive . s3://$(WEBUI_BUCKET) --acl public-read --exclude *settings.js
+  $(if $(filter none, $(WEBUI_BUCKET)), @echo "WebUI not deployed so no upload possible.", cd frontend/build && aws s3 cp --recursive . s3://$(WEBUI_BUCKET) --acl public-read --exclude *settings.js)
```
HieronymusLex, line 121: nit: indentation seems off?

docs/LOCAL_DEVELOPMENT.md
```diff
@@ -12,6 +12,9 @@ The following dependencies must be installed:
 - node.js >= v16.11 and npm >= 8
 - virtualenv
 - Ruby >= 2.6
+- libsnappy-dev/snappy-devel (debian/centos)
+- docker
```
HieronymusLex, line 16: which bit is docker for?

docs/UPGRADE_GUIDE.md
```diff
@@ -1,5 +1,28 @@
...
 
+## Migrating from <=v0.47 to v0.48
+
+If you are migrating to v0.48 and choosing to remove the WebUI component (before
+this version this option did not exist), you will receive errors in the
+CloudFormation updates. _This is expected._ There are 2 expected errors:
+
+- In the `DeployStack` the `CleanupWebUIBucket` resource will fail to delete.
+  This resource is just an attempt to empty the WebUI S3 Bucket which is no
+  longer possible as permissions to the bucekt have been removed.
```
HieronymusLex, line 11:
```suggestion
  longer possible as permissions to the bucket have been removed.
```

docs/UPGRADE_GUIDE.md
```diff
@@ -1,5 +1,28 @@
 # Upgrade Guide
 
+## Migrating from <=v0.47 to v0.48
+
+If you are migrating to v0.48 and choosing to remove the WebUI component (before
+this version this option did not exist), you will receive errors in the
+CloudFormation updates. _This is expected._ There are 2 expected errors:
+
+- In the `DeployStack` the `CleanupWebUIBucket` resource will fail to delete.
+  This resource is just an attempt to empty the WebUI S3 Bucket which is no
```
HieronymusLex, line 10:
```suggestion
  `CleanupWebUIBucket` is a custom resource which attempts to empty the WebUI S3 Bucket which is no
```

docs/UPGRADE_GUIDE.md
```diff
@@ -1,5 +1,28 @@
...
+  longer possible as permissions to the bucekt have been removed.
+
+- The `WebUIStack` stack will fail to delete. This is due to the S3 bucket not
+  being empty, which is due to the previous error.
+
+These errors can be very easily resolved:
```
HieronymusLex, line 16:
```suggestion
To resolve these errors:
```

docs/UPGRADE_GUIDE.md
```diff
@@ -1,5 +1,28 @@
...
+
+1. First wait until the main S3F2 stack is at `UPDATE_COMPLETE` status.
```
HieronymusLex, line 18:
```suggestion
1. Wait until the main S3F2 stack is at `UPDATE_COMPLETE` status.
```

docs/UPGRADE_GUIDE.md
```diff
@@ -1,5 +1,28 @@
...
+2. Locate the `WebUIStack` stack in the CloudFormation console and find the
+   WebUIBucket bucket name from looking at the Outputs tabs.
+3. Go to the S3 Console and select the WebUIBucket with the name found above,
```
HieronymusLex, line 21:
```suggestion
3. Navigate to the S3 Console and choose the WebUIBucket with the name found above,
```

docs/UPGRADE_GUIDE.md
```diff
@@ -1,5 +1,28 @@
...
+  This resource is just an attempt to empty the WebUI S3 Bucket which is no
+  longer possible as permissions to the bucekt have been removed.
+
+- The `WebUIStack` stack will fail to delete. This is due to the S3 bucket not
+  being empty, which is due to the previous error.
+
+These errors can be very easily resolved:
+
+1. First wait until the main S3F2 stack is at `UPDATE_COMPLETE` status.
+2. Locate the `WebUIStack` stack in the CloudFormation console and find the
```
HieronymusLex, line 19:
```suggestion
2. Choose the `WebUIStack` stack in the CloudFormation console and find the
```

docs/USER_GUIDE.md
```diff
@@ -163,6 +167,12 @@ resources.
      this stack in CloudFormation once deployed.
    - **AdminEmail:** The email address you wish to setup as the initial user of
      this Amazon S3 Find and Forget deployment.
+   - **DeployWebUI:** (Default: true) Choose whether to deploy the Web UI as
+     part of the solution. If you do deploy the Web UI AuthMethod must use
+     Cognito. If you do not use the Web UI interaction with the solution is done
+     through the APIs only.
```
HieronymusLex, line 173:
```suggestion
     part of the solution. If set to **true**, the AuthMethod parameter must be set
     to **Cognito**. If set to **false**, interaction with the solution is performed
     via the API Gateway only.
```

docs/USER_GUIDE.md
```diff
@@ -163,6 +167,12 @@ resources.
      this stack in CloudFormation once deployed.
    - **AdminEmail:** The email address you wish to setup as the initial user of
      this Amazon S3 Find and Forget deployment.
+   - **DeployWebUI:** (Default: true) Choose whether to deploy the Web UI as
```
HieronymusLex, line 170:
```suggestion
   - **DeployWebUI:** (Default: true) Whether to deploy the Web UI as
```

docs/USER_GUIDE.md
```diff
@@ -163,6 +167,12 @@ resources.
...
+     part of the solution. If you do deploy the Web UI AuthMethod must use
+     Cognito. If you do not use the Web UI interaction with the solution is done
+     through the APIs only.
+   - **AuthMethod:** (Default: Cognito) Choose which method of authentication
+     you would like to use. Cognito must be used if the Web UI is chosen.
```
HieronymusLex, line 175:
```suggestion
   - **AuthMethod:** (Default: Cognito) The authentication method to be
     used for the solution. Must be set to **Cognito** if DeployWebUI is true.
```

docs/USER_GUIDE.md
```diff
@@ -305,8 +316,13 @@ To add more users to the application:
 ### Making authenticated API requests
 
 To use the API directly, you will need to authenticate requests using the User
```
HieronymusLex, line 318: I don't think the bit about needing to use the user pool is correct given IAM is an option

docs/USER_GUIDE.md
```diff
@@ -305,8 +316,13 @@ To add more users to the application:
...
-Pool. After resetting the password via the UI, you can make authenticated
-requests using the AWS CLI:
+Pool. The method for authenticating differs depending on which authentication
+option was chosen.
```
HieronymusLex, line 320:
```suggestion
option was chosen:
```

docs/USER_GUIDE.md
```diff
@@ -330,6 +346,19 @@ requests using the AWS CLI:
    curl $API_URL/v1/queue -H "Authorization: Bearer $ID_TOKEN"
    ```
 
+#### IAM
+
+Unlike with the previous Cognito method, IAM authentication for API requests is
+done by performing the
+[Signature Version 4 signing process](https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html).
```
HieronymusLex, line 353:
```suggestion
IAM authentication for API requests uses the [Signature Version 4 signing process](https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html). Add the
resulting signature to the **Authorization** header when making requests to the API.
```

templates/api.yaml
```diff
@@ -79,17 +88,893 @@ Resources:
             HttpMethod: '*'
         - !Ref AWS::NoValue
       DefinitionBody:
-        'Fn::Transform':
-          Name: 'AWS::Include'
-          Parameters:
-            Location: ./api.definition.yml
+        openapi: "3.0.1"
```
HieronymusLex, line 91: oof, why the switch to have the API definition in the template? it makes this file massive. I assume it's something to do with conditions

templates/api.yaml
```diff
@@ -79,17 +88,893 @@ Resources:
...
+                - "Message"
+          responses:
+            InvalidRequest:
+              description: "Error response for invalid requests"
+              content:
+                application/json:
+                  schema:
+                    $ref: '#/components/schemas/Error'
+          securitySchemes:
+            CognitoAuthorizer:
```
HieronymusLex, line 961: probably should change this name as well because cognito might not be in use

docs/USER_GUIDE.md
```diff
@@ -163,6 +167,12 @@ resources.
      this stack in CloudFormation once deployed.
    - **AdminEmail:** The email address you wish to setup as the initial user of
      this Amazon S3 Find and Forget deployment.
+   - **DeployWebUI:** (Default: true) Choose whether to deploy the Web UI as
+     part of the solution. If you do deploy the Web UI AuthMethod must use
+     Cognito. If you do not use the Web UI interaction with the solution is done
+     through the APIs only.
+   - **AuthMethod:** (Default: Cognito) Choose which method of authentication
```
HieronymusLex, line 174: have you tested switching from cognito -> sigv4 and vice versa?

templates/template.yaml
```diff
@@ -136,15 +136,38 @@ Parameters:
     Description: Comma-delimited list of subnets to deploy Fargate tasks in. This parameter must be set if the DeployVpc parameter is "false", otherwise it is ignored.
     Type: CommaDelimitedList
     Default: ""
+  DeployWebUI:
```
HieronymusLex, line 139: keep params in alphabetical order for consistency.

templates/template.yaml
```diff
@@ -136,15 +136,38 @@ Parameters:
...
+    Description: Specify whether you would like a Web User Interface deployed. (AuthMethod must be Cognito if WebUI is deployed).
```
HieronymusLex, line 140: Try to keep wording like with the other params 
```suggestion
    Description: Deploy the Web User Interface for this solution.. (AuthMethod must be Cognito if WebUI is set to true).
```

templates/template.yaml
```diff
@@ -136,15 +136,38 @@ Parameters:
...
+    AllowedValues:
+      - "Cognito"
+      - "IAM"
+
+Rules:
+  ValidateAuth:
+    RuleCondition: !Equals [!Ref DeployWebUI, "true"]
+    Assertions:
+      - AssertDescription: IAM Auth cannot be chosen when deploying the WebUI.
+        Assert: !Not [!Equals [!Ref AuthMethod, "IAM"]]
```
HieronymusLex, line 159: in the interest of future proofing for other auth methods, should this check be checking that AuthMethod === Cognito?

templates/template.yaml
```diff
@@ -226,7 +249,7 @@ Resources:
       Parameters:
         AccessControlAllowOriginOverride: !Ref AccessControlAllowOriginOverride
         ConfigParameter: !Ref ConfigParameter
-        CognitoUserPoolArn: !GetAtt AuthStack.Outputs.CognitoUserPoolArn
+        CognitoUserPoolArn: !If [ShouldDeployCognito, !GetAtt AuthStack.Outputs.CognitoUserPoolArn, "none"]
```
HieronymusLex, line 252: I think convention is to supply an empty string rather than "none"

tests/acceptance/conftest.py
```diff
@@ -195,6 +200,20 @@ def request(
...
+            ] == "IAM":
+                creds = boto3.Session().get_credentials().get_frozen_credentials()
+                if "json" in kwargs:
+                    request = AWSRequest(
+                        method=method.upper(), url=url, data=json.dumps(kwargs["json"])
+                    )
+                else:
+                    request = AWSRequest(method=method.upper(), url=url)
+                SigV4Auth(
+                    creds, "execute-api", getenv("AWS_DEFAULT_REGION", "eu-west-1")
```
HieronymusLex, line 214: do we default to eu-west-1 anywhere else?

tests/acceptance/test_data_mappers.py
```diff
@@ -146,7 +152,9 @@ def test_it_creates_without_optionals(
     expected = deepcopy(data_mapper)
     expected["Format"] = "parquet"
     expected["CreatedBy"] = {
-        "Username": "aws-uk-sa-builders@amazon.com",
+        "Username": "aws-uk-sa-builders@amazon.com"
+        if stack["AuthMethod"] != "IAM"
+        else "N/A",
```
HieronymusLex, line 157: this is quite a significant change as CreatedBy is used for things like audit. Is there no way to determine who has done what when IAM is used?

HieronymusLex (changes requested):

docs/LOCAL_DEVELOPMENT.md
danjhd: The build process that is triggered with make deploy uses Docker to build the images.

templates/template.yaml
danjhd: Good point, however one of the next changes we plan is to add Cognito with mandatory MFA as an option. So at that point we would have more than one Cognito option so i choose this != way of checking.

tests/acceptance/conftest.py
danjhd: Yes, as you guessed I was using this default to be consistant, eu-west-1 is default option when we create the temp bucket.  This is found later in the same file (line 569).

tests/acceptance/test_data_mappers.py
danjhd: Good point, i'll investigate options for this....

templates/api.yaml
danjhd: Yep, I tried many different ways to do this. But the previous Includes tranform we were using does not support CFn conditions. It only supports !Ref and !GetAtt so my hand was forced. If you think of anything that would get around this please shout i am as "happy" as you in having to do this.

docs/USER_GUIDE.md
danjhd: Agreed

Makefile
danjhd: Agreed

templates/api.yaml
danjhd: Agreed

docs/USER_GUIDE.md
danjhd: I have added a Note on line 177 of this file now to address this.

templates/template.yaml
HieronymusLex: the implementation of that might not necessarily involve a new AuthMethod. for example it could be a separate param. even so you could use the !Or operator in that scenario

danjhd: Fair point, i had not considered doing it with !Or, thanks.

danjhd: Logic updated to match your suggestion

tests/acceptance/test_data_mappers.py
```diff
@@ -12,17 +12,21 @@
 
 
 @pytest.mark.auth
-def test_auth(api_client, data_mapper_base_endpoint):
+def test_auth(api_client, data_mapper_base_endpoint, stack):
     headers = {"Authorization": None}
+    status_code = 403 if stack["AuthMethod"] == "IAM" else 401
```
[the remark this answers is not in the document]
danjhd: I have moved the logic for this into a `fixture` so that the test no longer contains the condition. I know this is not truely what you meant. But first step.

Makefile
```diff
@@ -45,7 +44,8 @@ deploy-artefacts:
 .PHONY: format-cfn
 format-cfn:
 	$(eval VERSION := $(shell $(MAKE) -s version))
-	sed -i '' -e '3s/.*/Description: Amazon S3 Find and Forget \(uksb-1q2j8beb0\) \(version:$(VERSION)\)/' templates/template.yaml
+	sed -i.bak -e '3s/.*/Description: Amazon S3 Find and Forget \(uksb-1q2j8beb0\) \(version:$(VERSION)\)/' templates/template.yaml
```
ctd, line 47: Nitpick: you should use mktemp when a temporary filename is needed - otherwise this could overwrite/delete an existing file.

Makefile
```diff
@@ -72,15 +72,17 @@ format-python: | $(VENV)
 	; done
 
 generate-api-docs:
-	npx openapi-generator generate -i ./templates/api.definition.yml -g markdown -t ./docs/templates/ -o docs/api
+	cat ./templates/api.yaml | $(VENV)/bin/yq -y .Resources.Api.Properties.DefinitionBody > ./templates/temp.definition.yml
```
ctd, line 75: Nitpick: you should use mktemp when a temporary filename is needed - otherwise this could overwrite/delete an existing file.

backend/lambda_layers/boto_utils/python/boto_utils.py
```diff
@@ -213,12 +213,26 @@ def parse_s3_url(s3_url):
...
+            "Sub": claims.get("sub", "N/A"),
+        }
+    # If Cognito IAM is being used
+    elif "identity" in req:
+        iden = req.get("identity", {})
+        return {
+            "Username": iden.get("userArn", "N/A"),
+            "Sub": iden.get("user", "N/A"),
+        }
+    else:
```
ctd, line 231: Is this else-case an expected situation? (I'm thinking whether we should we be raising an exception/error rather than continuing with missing authn data?)

HieronymusLex: agreed - we should error here. I don't think there's any situation where we'd want to be able to perform operations without knowing whodunit

templates/template.yaml
```diff
@@ -136,15 +136,38 @@ Parameters:
...
+    RuleCondition: !Equals [!Ref DeployWebUI, "true"]
+    Assertions:
+      - AssertDescription: IAM Auth cannot be chosen when deploying the WebUI.
+        Assert: !Not [!Equals [!Ref AuthMethod, "IAM"]]
 
 Conditions:
   DefaultPreBuiltArtefactsBucket: !Equals [!Ref PreBuiltArtefactsBucketOverride, "false"]
   ShouldDeployVpc: !Equals [!Ref DeployVpc, "true"]
+  ShouldDeployWebUI: !Equals [!Ref DeployWebUI, "true"]
+  ShouldDeployCognito: !Not [!Equals [!Ref AuthMethod, "IAM"]]
```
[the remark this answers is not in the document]
danjhd: Yep, thanks i missed this one. Will change this one too

backend/lambda_layers/boto_utils/python/boto_utils.py
danjhd: Fair point. I'll make that change (channelling my inner MJ)

danjhd: Just committed a change to add an expection.

danjhd: I have reverted back to inckuding the N/A as the `else` block since we have a unit test already in place (before my changes) that sent an empty request request and expects N/A values back. So reverting back to this to keep the existing behaviour consistent

Makefile
danjhd: Thanks for the advice, i have done as suggested

Makefile
danjhd: Thanks for the advice, i have done as suggested

tests/acceptance/test_data_mappers.py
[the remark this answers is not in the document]
danjhd: I have just split the marks on the tests so there are now separate cognito and iam versions.

Makefile
```diff
@@ -72,15 +74,17 @@ format-python: | $(VENV)
 	; done
 
 generate-api-docs:
-	npx openapi-generator generate -i ./templates/api.definition.yml -g markdown -t ./docs/templates/ -o docs/api
+	$(eval TEMP_FILE := $(shell mktemp))
+	cat ./templates/api.yaml | $(VENV)/bin/yq -y .Resources.Api.Properties.DefinitionBody > $(TEMP_FILE)
```
ctd, line 78:
```suggestion
	$(VENV)/bin/yq -y .Resources.Api.Properties.DefinitionBody ./templates/api.yaml > $(TEMP_FILE)
```

Remove unnecessary use of cat :)

docs/USER_GUIDE.md
```diff
@@ -330,6 +355,20 @@ requests using the AWS CLI:
...
+Add the resulting signature to the **Authorization** header when making requests
+to the API.
+
+Use the Sigv4 process linked above to generate the Authorization header value
+and then call the API as normal:
+
+```sh
+curl $API_URL/v1/queue -H "Authorization: $Sigv4Auth"
+```
+
```
ctd, line 371: The line below this (372) should be under the Cognito section?

ctd (commented):
One more thing I think we need docs-wise is an overview of how I setup the solution using IAM auth. i.e. do I need to create an IAM user? role? what permissions control who/what can interact with s3f2?

Makefile
danjhd: Thank you. I also made one more change here where is did the same thing with TEMP_FILE variable as you suggested for the sed command

docs/USER_GUIDE.md
danjhd: Yep, thanks.

danjhd:
> One more thing I think we need docs-wise is an overview of how I setup the solution using IAM auth. i.e. do I need to create an IAM user? role? what permissions control who/what can interact with s3f2?

I have updated the Docs to include information on what is required from an IAM perspective.

templates/template.yaml
```diff
@@ -207,7 +232,6 @@ Resources:
       - LayersStack
       - StateMachineStack
       - StreamProcessorStack
-      - WebUIStack
```
matteofigus, line 210: should we keep the `WebUIStack` here with an IF for the web UI? I think the goal here is to avoid you receiving the email saying "Hey welcome to the thing, this is your user/temp password" before the UI stack is actually created (so the email having a url that is not accessible yet).

matteofigus (changes requested):
Well done, that's a lot of work :D I added a couple of questions.
Also, can we add some unit tests on the custom resources?

package-lock.json
```diff
@@ -39496,7 +39495,7 @@
       "resolved": "https://registry.npmjs.org/jake/-/jake-10.8.4.tgz",
       "integrity": "sha512-MtWeTkl1qGsWUtbl/Jsca/8xSoK3x0UmS82sNbjqxxG/de/M/3b1DntdjHgPMC50enlTNwXOCRqPXLLt5cCfZA==",
       "requires": {
-        "async": "3.2.2",
+        "async": "0.9.x",
```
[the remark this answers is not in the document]
danjhd: Thanks, fixed and commited the change.

danjhd:
> > Well done, that's a lot of work :D I added a couple of questions. Also, can we add some unit tests on the custom resources?
> 
> Yep, i realised on Monday that i missed the tests for the new CRs. I'll get that done too.

Unit tests for CRs just committed.

matteofigus (changes requested):
A minor issue and a question

templates/api.yaml
```diff
@@ -79,17 +88,893 @@ Resources:
...
+          All:
+            validateRequestParameters: true
+            validateRequestBody: true
       EndpointConfiguration: REGIONAL
-      Auth:
-        DefaultAuthorizer: CognitoAuthorizer
-        Authorizers:
-          CognitoAuthorizer:
-            UserPoolArn: !Ref CognitoUserPoolArn
-        AddDefaultAuthorizerToCorsPreflight: False
```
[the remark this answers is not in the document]
danjhd: As discussed offline. I have double checked and we do not need this code. The OPTIONS method is being deployed with None for Auth so it is as required and this change is not required.

docs/USER_GUIDE.md
```diff
@@ -332,6 +348,40 @@ requests using the AWS CLI:
...
+      "Effect": "Allow",
+      "Action": "execute-api:Invoke",
+      "Resource": "API_ARN"
+    }
+  ]
+}
+```
+
+**Note**: `API_ARN` is to be replaced with the value found in value of the
+`ApiUrl` CloudFormation Stack Output
```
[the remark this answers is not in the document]
danjhd: Good catch, thank you

matteofigus (approved):

backend/lambdas/custom_resources/redeploy_apigw.py
```diff
@@ -0,0 +1,31 @@
...
+    return None
+
+
+@with_logging
+@helper.update
+def update(event, context):
+    props = event["ResourceProperties"]
+    props_old = event["OldResourceProperties"]
+    if props_old["DeployCognito"] != props["DeployCognito"]:
+        api_client.create_deployment(
```
HieronymusLex, line 24: for my own understanding, what's the need to do this ourselves?

danjhd: It turns out that an update to change the auth of an an API does update the api but does not publish that new version so if the deployment is updated to change it this is required otherwise the auth of the API gets out of sync. New deployments work fine it is only the update with change that is affected.

matteofigus (approved):

HieronymusLex (approved):

PR #314, opened 2022-06-20
Updated build_manifest_row to generate queries based on the deletion queue type

adithyapathipaka:
*Description of changes:*
Updated build_manifest_row to generate queries based on the deletion queue type.

I had found an issue with the composite type deletion queue when the deletion job started, where the tasks will [generate Athena queries info](https://github.com/awslabs/amazon-s3-find-and-forget/blob/master/backend/lambdas/tasks/generate_queries.py#L88-L126) to check the existence of data, the **QueryableMatchId** is generated based on the no. of columns used in the deletion queue, but not based on the selection made in deletion queue, due to which when selecting one column with composite type manifest row generates match id based on simple type.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed


matteofigus:
Hi @adithyapathipaka and thank you for your PR. 

There are a couple of things needed for this PR:

1. Can you rebase from main branch and put the new feature under v0.49? We released 0.48 yesterday, something in the lines of 
```
## v0.49 (unreleased)

- [#314](https://github.com/awslabs/amazon-s3-find-and-forget/pull/314): Updated build_manifest_row to generate queries based on the deletion queue type
```
on top of the file?

2. Your code looks good, but are you able to create a unit test to cover this scenario? You can probably find a similar test that you can use as a base: https://github.com/awslabs/amazon-s3-find-and-forget/blob/master/tests/unit/tasks/test_generate_queries.py#L509-L588

If you can work on those changes it would be great, otherwise I can pull your changes into a new branch and work on those myself, just let me know.

Thanks again.

adithyapathipaka:
Hello @matteofigus,
Thanks for the reply, I have updated the CHANGELOG.md and test cases as well.

Regards,
Adithya

PR #318, opened 2022-06-23
China region support

danjhd:
*Issue #, if available:* n/a

*Description of changes:*
Updated the solution to be truely region independant so that it can be deployed into China region.

*PR Checklist:*

- [X] Changelog updated
- [X] Unit tests (and integration tests if applicable) provided
- [X] All tests pass
- [X] Pre-commit checks pass
- [X] Debugging code removed
- [X] If releasing a new version, have you bumped the version in the main CFN template?


matteofigus (approved):
LGTM

backend/lambdas/custom_resources/validate_vpce_subnets.py
```diff
@@ -0,0 +1,40 @@
...
+helper = CfnResource(json_logging=False, log_level="DEBUG", boto_level="CRITICAL")
+
+ec2_client = boto3.client("ec2")
+
+
+@with_logging
+@helper.create
+@helper.update
+def create(event, context):
+    props = event.get("ResourceProperties", None)
```
HieronymusLex, line 14: nit: if your code _needs_ these values to be set, then it's probably not worth using the `.get` method of `dict` because that implies you're fine with the default return value of `None` which it doesn't look like we are. In the event that one of these props isn't set, then presumably we want the function to error and rollback

I have seen this approach done in one of the other custom resources but I don't think it's correct, especially given this is a validation scenario. I've marked this comment as a "nit" because in reality, what would probably happen is that the function will error anyways when you try to call a method on `None`, but it will be slightly harder to debug

templates/vpc.yaml
```diff
@@ -1,6 +1,13 @@
 AWSTemplateFormatVersion: "2010-09-09"
+Transform: AWS::Serverless-2016-10-31
 Description: Amazon S3 Find and Forget VPC (uksb-1qjminsd5)
 
+Globals:
+  Function:
+    Runtime: python3.8
```
HieronymusLex, line 7: can we use 3.9?

templates/vpc.yaml
```diff
@@ -213,58 +225,136 @@ Resources:
...
+            Action:
+              - ec2:DescribeSubnets
+              - ec2:DescribeVpcEndpointServices
+            Resource: '*'
+
   # Endpoints
 
-  CloudWatchEndpoint:
-    Type: AWS::EC2::VPCEndpoint
+  CloudWatchEndpointSubnets:
```
HieronymusLex, line 244: Is there an additional endpoint being added here? If so the estimated costs need to be updated

tests/unit/crs/test_cr_validate_vpce_subnets.py
```diff
@@ -0,0 +1,80 @@
+import pytest
```
HieronymusLex, line 1: ideally there'd be a negative test somewhere in here for unexpected input

HieronymusLex (changes requested):
are there any gotchas or specific things that need to be called out in the README for these changes?

HieronymusLex:
In terms of "gotachas" or notable differences that need calling out, the main one I can think of is around the WebUI/Auth

backend/lambdas/custom_resources/validate_vpce_subnets.py
```diff
@@ -0,0 +1,40 @@
...
+
+helper = CfnResource(json_logging=False, log_level="DEBUG", boto_level="CRITICAL")
+
+ec2_client = boto3.client("ec2")
+
+
+@with_logging
+@helper.create
+@helper.update
+def create(event, context):
```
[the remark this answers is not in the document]
danjhd: There are some services that do not have VPCEs in China region and also some VPCEs for some services do not support all AZs. Therefore we needed a CR to return valid subnets. Also the domain name of the service varies depending on the service so we needed the CR to return the correct name too.

backend/lambdas/custom_resources/validate_vpce_subnets.py
danjhd: Yes, i agree. As you said there are other CRs that do it in this project so i did it here to follow the others and be consistant.

templates/vpc.yaml
danjhd: Yes, we can and we should. I'll change it.

templates/vpc.yaml
danjhd: No, there are no new endpoints being added. They are the same as before this PR.

backend/lambdas/custom_resources/validate_vpce_subnets.py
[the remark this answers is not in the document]
matteofigus: I am not a big fan of comments in code but I think in this particular scenario, perhaps we can add a note on top of this file to explain this. Just to make sure a reader can have the necessary context to understand the reason this exists?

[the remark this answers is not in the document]
HieronymusLex: fair enough, we should probably use a different name then because the goal of the CR isn't really validation, it's to retrieve some data if I've understood correctly

HieronymusLex (dismissed):

backend/lambdas/custom_resources/validate_vpce_subnets.py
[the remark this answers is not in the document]
danjhd: Name of function changed and note added to top of file to help

tests/unit/crs/test_cr_get_vpce_subnets.py
```diff
@@ -0,0 +1,125 @@
...
+        ]
+    }
+
+    mock_client.describe_vpc_endpoint_services.return_value = {"ServiceDetails": []}
+
+    with pytest.raises(Exception) as e_info:
+        create(event, MagicMock())
+
+    mock_client.describe_subnets.assert_called_with(SubnetIds=[])
+    mock_client.describe_vpc_endpoint_services(
```
matteofigus, line 98: Should this be this instead?

```suggestion
    mock_client.describe_vpc_endpoint_services.assert_called_with(
```

PR #322, opened 2022-06-24
Python3.9

danjhd:
*Issue #, if available:* n/a

*Description of changes:*
Updated Python to 3.9 throughout (requirements regenerated as a result).

*PR Checklist:*

- [X] Changelog updated
- [X] Unit tests (and integration tests if applicable) provided
- [X] All tests pass
- [X] Pre-commit checks pass
- [X] Debugging code removed
- [X] If releasing a new version, have you bumped the version in the main CFN template?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #327, opened 2022-07-20
Submit Query Lambda Performance improvements

matteofigus:
*Description of changes:*

Improve memory management for the Submit Query Lambda. The current work should be useful when receiving hundreds of thousands of results for a single query. Rather than reading the whole set of results and storing in memory, with this code changes we read in batches using the generator that makes paginated calls to API, and then we save results to SQS earlier and in batches in order to use less memory.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed

### DIFF
--- a/backend/ecs_tasks/delete_files/utils.py
+++ b/backend/ecs_tasks/delete_files/utils.py
@@ -3,7 +3,7 @@ from botocore.exceptions import ClientError
 
 
 def remove_none(d: dict):
-    return {k: v for k, v in d.items() if v is not None and v is not ""}
+    return {k: v for k, v in d.items() if v is not None and v != ""}
 
 
 def retry_wrapper(fn, retry_wait_seconds=2, retry_factor=2, max_retries=5):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change Log
 
+## v0.53 (unreleased)
+
+- [#327](https://github.com/awslabs/amazon-s3-find-and-forget/pull/327): Improve
+  performance of Athena results handler
+
 ## v0.52
 
 - [#318](https://github.com/awslabs/amazon-s3-find-and-forget/pull/318): Added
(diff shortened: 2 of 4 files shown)


backend/ecs_tasks/delete_files/utils.py
```diff
@@ -3,7 +3,7 @@
 
 
 def remove_none(d: dict):
-    return {k: v for k, v in d.items() if v is not None and v is not ""}
+    return {k: v for k, v in d.items() if v is not None and v != ""}
```
matteofigus, line 6: This is slightly unrelated, fixes a warning that occurs when running unit tests, introduced when we migrated to Python 3.9 in #322 

```
backend/ecs_tasks/delete_files/utils.py:6
  /home/runner/work/amazon-s3-find-and-forget/amazon-s3-find-and-forget/backend/ecs_tasks/delete_files/utils.py:6: SyntaxWarning: "is not" with a literal. Did you mean "!="?
    return {k: v for k, v in d.items() if v is not None and v is not ""}
-- Docs: https://docs.pytest.org/en/stable/warnings.html
```

HieronymusLex (approved):
Looks good - do you have any data on performance gains?

backend/ecs_tasks/delete_files/utils.py
danjhd: I had a wanring on this when I did the upgrade to 3.9, I should have changed it then, sorry.

danjhd (approved):
LGTM

matteofigus:
> Looks good - do you have any data on performance gains?

I did try to make a real integration test with 5000 results, the difference is quite small:

Old code: `Duration: 17561.18 ms	Billed Duration: 17562 ms	Memory Size: 512 MB	Max Memory Used: 86 MB	Init Duration: 847.30 ms`

New: `Duration: 17539.00 ms	Billed Duration: 17539 ms	Memory Size: 512 MB	Max Memory Used: 84 MB	Init Duration: 1250.38 ms`

I think at this size, the improvement is small (only 2MB difference, 3% less memory usage) but when growing to hundreds of thousands, that is better than nothing!

PR #332, opened 2022-07-28
Remove S3FS dependency

matteofigus:
*Description of changes:*

This PR removes s3fs to simplify handling communications with S3.

It uses [pyarrow's S3FileSystem](https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html) to read, and [boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.upload_fileobj) to write back.

The reason I am using pyarrow to read is that it allows us to get rid of s3fs and I haven't noticed differences in terms of performance.

I tried to use pyarrow to write back to s3 too, but unfortunately I don't have an API that gives me back the VersionId of the write. If I would read the latest version using boto3, I would lose confidence that the version I'm reading is the one I just wrote, and therefore I wouldn't be able to perform consistency checks.

In order to remove s3fs, I investigated boto3 and found a (new?) convenient method called [upload_fileobj](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.upload_fileobj) that is smart enough to figure out if doing multipart or not, and I noticed an increase in performance compared to s3fs. The issue is that this method doesn't return the versionID as well, but I managed to find a way to monkey patch the code while we wait for [this issue](https://github.com/boto/s3transfer/issues/82#issuecomment-837971614) to be resolved. I found some Pull Requests too so perhaps we can see this soon landing in boto.

I ran the acceptance tests as well as some manual tests with very large files (8GB snappy compressed parquet), for which I didn't see any difference in reads, and some performance improvement (3x) in writing.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed


backend/ecs_tasks/delete_files/main.py
```diff
@@ -141,21 +144,26 @@ def execute(queue_url, message_body, receipt_handle):
...
+            role_arn=body.get("RoleArn"),
+            load_frequency=60 * 60,
         )
         # Download the object in-memory and convert to PyArrow NativeFile
         logger.info("Downloading and opening %s object in-memory", object_path)
-        metadata = s3.metadata(object_path, refresh=True)
-        with s3.open(object_path, "rb") as f:
-            source_version = f.version_id
+        with s3.open_input_stream(
+            "{}/{}".format(input_bucket, input_key), buffer_size=FIVE_MB
```
matteofigus, line 157: Buffer size is same as we used to do it with s3fs: https://github.com/fsspec/s3fs/blob/main/s3fs/core.py#L246

tests/unit/ecs_tasks/test_s3.py
```diff
@@ -725,3 +730,46 @@ def test_it_caches_manifests(mock_fetch):
...
+    when we write back to S3. The issue is that while the method offers a nice
+    wrapper around file operations, such as implementing multipart upload only
+    when needed, we don't get the VersionId back as we would by using
+    put_object. This monkey patch is in place while we wait some pull requests
+    to be merged. In the meanwhile, here is a test that allow us to notice
+    any change on the version we use on s3transfer, in order to add extra
+    protection against automated library upgrade PRs that may silently introduce
+    issues.
+    """
+    assert s3transfer.__version__ == "0.6.0"
```
matteofigus, line 749: I discussed this with @HieronymusLex - may seem a bit over cautious but I think it's worth it given Dependabot may make very easy to silently upgrade and leave dangerous things unnoticed.

HieronymusLex (approved):

HieronymusLex (approved):

PR #348, opened 2022-12-20
Remove redundant put call to DDB

matteofigus:
There is a duplicated call to put_item in DDB that has no effect in functionality (put for same ID) but potentially duplicates cost and performance for adding items to the queue. This PR is to remove this redundant call.

It is caused by a mistake when doing the refactoring in https://github.com/awslabs/amazon-s3-find-and-forget/pull/200

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change Log
 
+## v0.57 (unreleased)
+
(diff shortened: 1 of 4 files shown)


ctd (approved):
LGTM

Issue #361, opened 2023-04-26
Security Vulnerabilities found in ECR repository


abhra-gupta-trakstar:
There are 4 Critical Vulnerabilities our Third Party Scanning Tool has exposed in the ECR Repository of the S3F2 Deletion Stack.
These seem to have been introduced by the base image (python:3.9.13)
**I am using S3F2 v:0.53**
Attached are the vulnerabilities which are classified as `Critical` by Snyk.
I have little to no idea where to start fixing the issue from. 
If you could point to a right direction in fixing these, that would be great!

[s3f2-delstack-109grp733tnhy-ecrrepository-asnjsln0ge7n_latest _ Snyk.pdf](https://github.com/awslabs/amazon-s3-find-and-forget/files/11331121/s3f2-delstack-109grp733tnhy-ecrrepository-asnjsln0ge7n_latest._.Snyk.pdf)

ctd:
I believe this issue should be resolved by updating to v0.59 which was released today.

Please see the [updating instructions](https://github.com/awslabs/amazon-s3-find-and-forget/blob/master/docs/USER_GUIDE.md#updating-the-solution) for how to update your installation.

You can reopen this issue if you have any outstanding questions or concerns.

PR #362, opened 2023-04-26
Improvements for container image build process

ctd:
*Issue #, if available:* https://github.com/awslabs/amazon-s3-find-and-forget/issues/361

*Description of changes:*

This change:
* Removes bundling the container base image with a release
* Uses the public ECR repository to retrieve the base image
* Specifies a label to pin the base image to Debian 11.6 "bullseye"
* Adds a build step to upgrade Debian packages

Since bundling the base image was added in dd7438c, ECR Public has become an option to reliably retrieve the base image. This saves us from having to bundle it in the release artefacts and for deployments to be pinned to the version at-release-time.

This, in combination with adding a package upgrade step the Dockerfile, makes it possible to rebuild the container image to update system libraries when security updates become available from the upstream image publisher/operating system vendor.

To mitigate compatibility concerns, I have pinned the image to the Debian 11.6 "bullseye" stable release. This should mean we get security patches automatically, but not breaking version changes.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed


backend/ecs_tasks/delete_files/Dockerfile
```diff
@@ -1,9 +1,9 @@
 ARG src_path=backend/ecs_tasks/delete_files
 ARG layers_path=backend/lambda_layers
 
-FROM python:3.9-slim as base
+FROM public.ecr.aws/docker/library/python:3.9-slim-bullseye as base
```
matteofigus, line 4: I understand bullseye could be a version available but not declared stable yet, why do we want that?

matteofigus (commented):
Thanks for working on this.

I am in favour of moving from Docker Hub to AWS for fetching the base image.

Nevertheless, I am not sure about the dynamic fetching of the image during build time. I can see the benefit of allowing an updated fetch + upgrading of dependencies, but I think that comes with the cost of not allowing deterministic builds. That is because the `python:3.9-slim-bullseye` image gets updated over time during build rather than release (and different fetches may use a different image that uses a different versions of python), and the update command will update other dependencies too. There are 2 main consequences I can think of:
* A customer deployment has higher changes to use a different environment than the one used to run Acceptance Tests (but this risk is already present as the fetch of the image is currently done on merge/release, while acceptance run on developer's machine during development - still, while one may be days, now it could become months or more). This could mean higher changes to introduce bugs, and make hard for us to support them during troubleshooting, as it would rely on them finding out, for example, which version of Python they are running, rather than looking at the version of Python that we know is pinned to a particular release.
* Multi-environment deployments (for example, Test and Prod) may have substantial differences such as version of Python running the code. This is an anti-pattern as it affects tests reliability, including manual and security related that the customer may have in place.

Having said that, I am still forming an opinion and I'd like to think about this a little more. I see the benefit of simplifying this code and running faster releases (as we decrease the packaged code during build).

Perhaps the best way to mitigate some of my concerns would be to at least lock the patch of Python by switching from `3.9-slim` to `3.9.16-slim` - at least we are locking the python version.

What do you think?

backend/ecs_tasks/delete_files/Dockerfile
ctd: "[bullseye](https://www.debian.org/releases/bullseye/)" refers to the Debian release that the image is built on. It is the current "[stable](https://wiki.debian.org/Status/Stable)" version of Debian.

ctd: The reason we would want to pin to a particular Debian release is that the stable release has minimal changes in terms of updates.

ctd:
Agree with pinning the python version, I'll revise and add that in.

In terms of deterministic builds, we don't really have those as-is because the versions of `g++ gcc libsnappy-dev` are downloaded/installed at build-time. That said, I'm fairly confident by pinning the Debian release version that this is less of a concern.

ctd:
I've updated the Dockerfile base image to be 3.9.16 specific.

matteofigus (approved):

PR #367, opened 2023-06-08
Fix codepipeline build

matteofigus:
*Issue #, if available:*

Starting May 25, 2023, CodePipeline doesn't allow anymore polling based S3 sources. New stacks don't poll, and existing stacks with pipelines that haven't ran for more than 30 days, don't start a build when a S3 source is uploaded. As a consequence:
1. New builds of this solution are currently stuck on the "Wait for Container Build" custom resource, as the build never starts
2. Existing deployments of this solution cannot be updated for the same issue.

*Description of changes:*

There is [a (quite convoluted) way to setup Eventbridge to start a deployment, based on CloudTrail and setting up lots of resources](https://docs.aws.amazon.com/codepipeline/latest/userguide/update-change-detection.html#update-change-detection-S3), but we already have a Custom Resource Lambda to start the build, previously used to update stacks whenever the "Deploy Frontend" setting changed on top of an existing deployment. So, I opted to change it to be executed on stack creation as well as updates, and whenever a new version is published. 

I tested this change by updating an existing stack, as well as creating a new stack.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed

### DIFF
--- a/backend/lambdas/custom_resources/rerun_pipeline.py
+++ b/backend/lambdas/custom_resources/rerun_pipeline.py
@@ -19,12 +19,19 @@ def delete(event, context):
 @helper.update
 def update(event, context):
     props = event["ResourceProperties"]
-    props_old = event["OldResourceProperties"]
+    props_old = event.get("OldResourceProperties", {})
 
     deploy_ui_changed = (
-        props_old["DeployWebUI"] == "false" and props["DeployWebUI"] == "true"
+        "DeployWebUI" in props_old
+        and props_old["DeployWebUI"] == "false"
+        and props["DeployWebUI"] == "true"
     )
-    new_version = not "Version" in props_old or props_old["Version"] != props["Version"]
+
+    new_version = (
+        (not "Version" in props_old or props_old["Version"] != props["Version"])
+        if "Version" in props
+        else "Version" in props_old
+    )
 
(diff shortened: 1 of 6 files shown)


svozza (approved):

backend/lambdas/custom_resources/rerun_pipeline.py
```diff
@@ -9,19 +9,33 @@
...
-    if props_old["DeployWebUI"] == "false" and props["DeployWebUI"] == "true":
+    props_old = event.get("OldResourceProperties", {})
+
+    deploy_ui_changed = (
+        "DeployWebUI" in props_old
+        and props_old["DeployWebUI"] == "false"
+        and props["DeployWebUI"] == "true"
+    )
+
+    new_version = (
```
svozza, line 30: i find this a little hard to read, i think a normal if/else rather than a ternery might be clearer.

matteofigus: Cool. I see you approved so I'll go merge now, but happy to revisit in a separate PR.

svozza: nah, it's just a personal thing, don't worry

PR #396, opened 2024-01-30
Improve JSON processing performance

matteofigus:
*Description of changes:*

Follow-up PR from #395 - move the optimisation logic outside of the JSON handler to simplify logic and use less memory (as the hashmap is created at the start of the job, rather than copied on each line iterator).

I also included File Size in the log as it could be useful during troubleshooting.

*PR Checklist:*

- [x] Changelog updated
- [x] All tests pass
- [x] Pre-commit checks pass
- [x] Debugging code removed

### DIFF
--- a/backend/ecs_tasks/delete_files/parquet_handler.py
+++ b/backend/ecs_tasks/delete_files/parquet_handler.py
@@ -114,14 +114,19 @@ def cast_column_values(column, schema):
         if is_column_type_decimal(schema, column["Column"]):
             column["MatchIds"] = set(Decimal(m) for m in column["MatchIds"])
     else:
-        for i in range(0, len(column["Columns"])):
-            if is_column_type_decimal(schema, column["Columns"][i]):
-                columns_copy = set()
-                for composite_match_tuple in column["MatchIds"]:
-                    match_array = [x for x in composite_match_tuple]
+        decimal_columns = [
+            i
+            for i in range(0, len(column["Columns"]))
+            if is_column_type_decimal(schema, column["Columns"][i])
+        ]
+        if len(decimal_columns):
+            columns_copy = set()
+            for composite_match_tuple in column["MatchIds"]:
+                match_array = [x for x in composite_match_tuple]
+                for i in decimal_columns:
                     match_array[i] = Decimal(match_array[i])
-                    columns_copy.add(tuple(match_array))
-                column["MatchIds"] = columns_copy
+                columns_copy.add(tuple(match_array))
+            column["MatchIds"] = columns_copy
     return column
 
 
--- a/backend/ecs_tasks/delete_files/parquet_handler.py
+++ b/backend/ecs_tasks/delete_files/parquet_handler.py
@@ -114,19 +114,19 @@ def cast_column_values(column, schema):
         if is_column_type_decimal(schema, column["Column"]):
             column["MatchIds"] = set(Decimal(m) for m in column["MatchIds"])
     else:
-        decimal_columns = [
+        decimal_columns = set(
             i
-            for i in range(0, len(column["Columns"]))
-            if is_column_type_decimal(schema, column["Columns"][i])
-        ]
-        if len(decimal_columns):
-            columns_copy = set()
-            for composite_match_tuple in column["MatchIds"]:
-                match_array = [x for x in composite_match_tuple]
-                for i in decimal_columns:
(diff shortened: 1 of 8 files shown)


backend/ecs_tasks/delete_files/parquet_handler.py
```diff
@@ -114,12 +112,16 @@ def cast_column_values(column, schema):
...
     if column["Type"] == "Simple":
         if is_column_type_decimal(schema, column["Column"]):
-            column["MatchIds"] = [Decimal(m) for m in column["MatchIds"]]
+            column["MatchIds"] = set([Decimal(m) for m in column["MatchIds"]])
     else:
         for i in range(0, len(column["Columns"])):
             if is_column_type_decimal(schema, column["Columns"][i]):
-                for composite_match in column["MatchIds"]:
-                    composite_match[i] = Decimal(composite_match[i])
+                columns_copy = set()
```
cmclel7, line 119: I'm wondering if there is a way to do this without creating another copy of the column.

Do you think that there is a possibility for there to be particularly large columns here? 

Primarily driven by the fact we have seen some large simple match lists but I haven't seen any use of composite personally.

backend/ecs_tasks/delete_files/parquet_handler.py
```diff
@@ -114,12 +112,16 @@ def cast_column_values(column, schema):
     """
     if column["Type"] == "Simple":
         if is_column_type_decimal(schema, column["Column"]):
-            column["MatchIds"] = [Decimal(m) for m in column["MatchIds"]]
+            column["MatchIds"] = set([Decimal(m) for m in column["MatchIds"]])
```
ctd, line 115:
```suggestion
            column["MatchIds"] = set(Decimal(m) for m in column["MatchIds"])
```

You can skip creating an intermediate list by omitting the braces.

backend/ecs_tasks/delete_files/parquet_handler.py
matteofigus: The possibility is real but composite are definitely uncommon, yes, and also this is effectively running if a column identifier is also of type decimal, which is very uncommon, so I think the possibility of hitting both use-cases is very remote.

But your comment is valid. I guess before we were operating on arrays, and therefore re-iterating and casting the Decimals in this particular use-case wasn't memory intensive because we were operating on the existing data structure. I guess here we are sort of doing the opposite really, working out the hashmaps before, and copying to array only here.

I guess that if you have more than a decimal column identifier, this could be doing the copy multiple times, and perhaps I can optimise that to happen only once. I'll give it a go.

matteofigus: I did re-write some logic to optimise the multi-column decimal scenario. I think this is the best I can think of, because I don't think iterating over a Python set allows me to dynamically add/remove values? I think the best is to just create a copy and re-write it at the end. Thoughts @ctd @cmclel7 ?

ctd: Yeah, I think you need to build a new collection.

You could use generator expressions here as well, for instance:
```python
def cast_column_values(column, schema):
    """
    Method to cast stringified MatchIds to their actual types
    """
    if column["Type"] == "Simple":
        if is_column_type_decimal(schema, column["Column"]):
            column["MatchIds"] = set(Decimal(m) for m in column["MatchIds"])
    else:
        decimal_columns = set(
            i
            for i, col in enumerate(column["Columns"])
            if is_column_type_decimal(schema, col)
        )
        if decimal_columns:
            decimal_casted = set(
                tuple(
                    Decimal(m) if i in decimal_columns else m
                    for i, m in enumerate(composite_match_tuple)
                )
                for composite_match_tuple in column["MatchIds"]
            )
            column["MatchIds"] = decimal_casted
    return column
```

tbf after writing this I'm not sure that it will actually perform better. It saves having to build an intermediate list, but rather than iterating over decimal_columns (and only looping that many times) it adds a check step to len(composite_match_tuple). Swings and roundabouts perhaps.

matteofigus: Ah I actually like your changes. Using a set on the `decimal_columns` will have slight impact given I would assume the array to be usually around 2 or 3 but if we consider doing that lookup to up to millions, it could help.

I also like the idea of not using an intermediate array for `match_array`, so I incorporated this in my change. Thanks!

PR #453, opened 2026-06-08
fix(json): delete records with falsy match identifiers (0, "", false)

valter-silva-au:
## Problem

The JSON Forget-phase handler (`delete_matches_from_json_file`) silently fails to delete records whose match identifier is a legitimate falsy JSON value — empty string `""`, integer `0`/float `0.0`, or boolean `false`. The matching record is written back to the output instead of being removed.

**The Composite case is the one that matters in practice.** Composite matches exist for datasets where no single column is unique, so the tuple components are ordinary attributes where falsy values are entirely plausible (`age: 0`, an empty optional name field). A falsy component was dropped from the reconstructed `matched` tuple, producing a malformed shorter tuple that no longer equals any entry in `MatchIds` — so a record that genuinely matches a deletion request is not erased.

The Simple case is the same bug but largely theoretical: Simple Match IDs are personal identifiers, where a real erasure request for `userid=0` or `""` isn't a scenario we realistically expect. The code was still wrong there (and inconsistent with the Parquet handler, which casts and compares falsy values correctly), so the fix covers it too.

## Root cause

Both comparison sites used Python truthiness rather than an explicit presence check:

- **Composite:** `if record:` — a falsy component is dropped from the `matched` list, so the reconstructed tuple no longer equals any entry in `MatchIds`.
- **Simple:** `if record and record in column["MatchIds"]:` — the `record and` clause short-circuits whenever `record` is falsy-but-present, so `0`/`""`/`false` never reach the membership test.

The `None` sentinel returned by `get_value`/`find_key` means **only** that a key segment is **absent**. So `is not None` is the precise predicate: it preserves the original intent (skip absent keys) while correctly admitting falsy-but-present values. This is the absent-vs-falsy distinction the truthiness guard conflated. (The Parquet path casts column values before comparison; the JSON path passes Match IDs through with native parsed types, which is why falsy natives leak into the guard here.)

## Fix

Two one-token changes:

- `if record:` → `if record is not None:`
- `if record and record in column["MatchIds"]:` → `if record is not None and record in column["MatchIds"]:`

## Testing

Added 4 unit tests in `tests/unit/ecs_tasks/test_json.py` — a Composite match with a `0` component (the case that matters), plus Simple `0`, Simple `""`, and Simple `false`; each asserts the matching record IS deleted. They fail on `master` (record survives, `DeletedRows: 0`) and pass after the fix. Full `test_json.py` (18) and the entire `tests/unit/ecs_tasks/` suite pass, including the existing nullable/undefined tests that pin absent-key behavior.

> The Simple `false` test is defensive coverage only — `boolean` is not in `ALLOWED_TYPES`, so a boolean column can never be selected as an identifier (`cast_to_type` would reject it upstream). The test is annotated to say so; it pins the handler's `is not None` semantics independent of the type rules.

## Risk

Low and tightly scoped. The only behavior change is that falsy-but-present identifiers now match. Absent keys still short-circuit (still `None`, still skipped) — guarded by pre-existing tests. No public API or signature change.

> No issue to close — this is an unreported bug found by reading the code. An alternative fix would align the JSON path with the Parquet path's explicit type-casting; I chose the minimal `is not None` guard as it directly targets the defect — flag if you'd prefer the casting approach.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Change Log
 
+## Unreleased
+
+- Fix the JSON Forget-phase handler so that records whose match identifier is a
+  legitimate falsy JSON value (integer `0`, float `0.0`, empty string `""`, or
+  boolean `false`) are correctly deleted. Previously a truthiness guard
+  short-circuited on these values and silently skipped matching records, which
+  is a GDPR right-to-be-forgotten correctness defect. The guard now uses
+  `is not None` to distinguish an absent key from a falsy-but-present value.
+
 ## v0.76 (2026-07-07)
 
 - [#457](https://github.com/awslabs/amazon-s3-find-and-forget/pull/457): Bump
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,10 @@
 
 ## Unreleased
 
-- Fix the JSON Forget-phase handler so that records whose match identifier is a
-  legitimate falsy JSON value (integer `0`, float `0.0`, empty string `""`, or
-  boolean `false`) are correctly deleted. Previously a truthiness guard
-  short-circuited on these values and silently skipped matching records, which
-  is a GDPR right-to-be-forgotten correctness defect. The guard now uses
-  `is not None` to distinguish an absent key from a falsy-but-present value.
+- [#453](https://github.com/awslabs/amazon-s3-find-and-forget/pull/453): Fix the
+  JSON Forget-phase handler so that Composite match components (and Simple match
+  IDs) with a falsy value such as `0` or `""` are no longer skipped by the
+  truthiness guard; the comparison now uses `is not None`.
 
 ## v0.76 (2026-07-07)
 
--- a/backend/ecs_tasks/delete_files/json_handler.py
+++ b/backend/ecs_tasks/delete_files/json_handler.py
@@ -52,14 +52,14 @@ def delete_matches_from_json_file(input_file, to_delete, compressed=False):
             for column in to_delete:
                 if column["Type"] == "Simple":
                     record = get_value(column["Column"], parsed)
-                    if record and record in column["MatchIds"]:
+                    if record is not None and record in column["MatchIds"]:
                         should_delete = True
                         break
                 else:
                     matched = []
                     for col in column["Columns"]:
                         record = get_value(col, parsed)
-                        if record:
+                        if record is not None:
                             matched.append(record)
                     if tuple(matched) in column["MatchIds"]:
                         should_delete = True
--- a/tests/unit/ecs_tasks/test_json.py
+++ b/tests/unit/ecs_tasks/test_json.py
@@ -366,6 +366,12 @@ def test_delete_correct_rows_from_json_file_with_falsy_empty_string_identifier()
 
 
 def test_delete_correct_rows_from_json_file_with_falsy_boolean_identifier():
+    # Defensive coverage only: `boolean` is not in ALLOWED_TYPES
+    # (generate_queries.py), so a boolean column can never be selected as a
+    # Match ID identifier and cast_to_type would reject it upstream. This test
+    # pins the handler's falsy-value behavior directly, independent of that
+    # guard, so the `is not None` semantics stay correct if the type rules
+    # ever change.
     # Arrange
     to_delete = [{"Column": "opted_in", "MatchIds": set([False]), "Type": "Simple"}]
     data = (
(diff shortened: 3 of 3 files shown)


matteofigus:
Thanks for the contribution — the analysis in the description is accurate and the fix is correct. I verified locally: the new tests fail on `master` and pass with the change, the full `tests/unit/ecs_tasks` suite passes, and the `is not None` guard preserves the absent-key behavior pinned by the existing nullable/undefined tests.

A few notes on framing and scope before this can go in:

1. **The Simple-match cases are mostly theoretical.** Match IDs in this project are personal identifiers, and a real erasure request for `userid=0` or an empty string is not a scenario we realistically expect. The code was still wrong (and inconsistent with the Parquet handler, which casts and compares falsy values correctly), so the fix is welcome — but I'd ask you to tone down the "GDPR compliance defect" framing in the description and CHANGELOG accordingly.

2. **The Composite case is the one that matters.** Composite matches exist for datasets where no single column is unique, so tuple components are ordinary attributes where falsy values are entirely plausible (`age: 0`, empty name fields). The old `if record:` also silently dropped a present component, producing a malformed shorter tuple for comparison. I'd suggest leading with this case in the PR description and CHANGELOG entry.

3. **The boolean `false` test pins unreachable behavior.** `boolean` isn't in `ALLOWED_TYPES` in `generate_queries.py`, so a boolean column can never be an identifier and `cast_to_type` would reject it upstream. I'm fine keeping the test as defensive coverage, but please add a brief comment in the test noting this, or drop it.

4. **CHANGELOG**: please trim the entry to a single factual line following the existing style, including the PR link (`[#453](...)`).

I'll run the full acceptance test suite on my side before merging. Happy to take this once the description/CHANGELOG tweaks are in.

---

*This code review was done by Claude Fable 5 and reviewed by @matteofigus.*

matteofigus (changes requested):
as explained in the comment

valter-silva-au:
Thanks for the careful review @matteofigus — all four addressed:

1. **Framing.** Reworked the PR description to lead with the **Composite** case (where falsy attribute values like `age: 0` are genuinely plausible) and dropped the "GDPR compliance defect" language; the Simple case is now described as the same bug but largely theoretical.

2. **Composite leads.** Both the description and the CHANGELOG entry now lead with Composite.

3. **Boolean `false` test.** Kept as defensive coverage with a comment noting `boolean` isn't in `ALLOWED_TYPES`, so a boolean column can never be an identifier and `cast_to_type` rejects it upstream — the test pins the handler's `is not None` semantics independent of that guard.

4. **CHANGELOG.** Trimmed to a single factual line in the existing style, including the PR link `[#453](...)`.

No change to `json_handler.py` — the `is not None` fix is unchanged. `test_json.py` (18) and `tests/unit/ecs_tasks/` stay green locally. Ready for the acceptance run whenever suits.

valter-silva-au:
Thanks for the thorough review, @matteofigus — all four points are addressed and pushed:

1. **Framing toned down** — dropped the "GDPR compliance defect" language from both the description and CHANGELOG; the Simple case is now described as largely theoretical.
2. **Lead with Composite** — the description and CHANGELOG entry now lead with the Composite case (falsy tuple components producing a malformed shorter tuple), with Simple noted as the same bug but theoretical.
3. **Boolean test annotated** — added a comment noting `boolean` isn't in `ALLOWED_TYPES`, so it's defensive coverage pinning the `is not None` semantics rather than a reachable path.
4. **CHANGELOG trimmed** — single factual line in the existing style, with the `[#453]` link.

Also rebased onto the latest `master` so it's current (CI re-running now). Ready for re-review whenever you have a moment — happy for you to run the acceptance suite on your side.

valter-silva-au:
Rebased onto `master` (now at v0.76) to clear a CHANGELOG merge conflict the recent release bumps introduced — the branch is `MERGEABLE` again, head `25bc97d`. The fix itself is unchanged and the four points from your review are still addressed (Composite-case-first framing, GDPR wording toned down, one-line CHANGELOG entry, boolean test annotated as defensive). Unit tests green (18/18). Happy to adjust anything else — ready for another look when you have a moment.

valter-silva-au:
Gentle nudge on this one, @matteofigus — it's been ready since the rebase on 8 Jul.

Current state: even with `master` (no conflicts), head `25bc97d`, unit tests green. All four points from your review are addressed — Composite-case-first framing, GDPR wording toned down, one-line CHANGELOG entry, and the boolean test annotated as defensive coverage (since `boolean` isn't in `ALLOWED_TYPES`).

The review still shows as changes-requested, which I think is just leftover state from before those fixes landed. Would you be able to take another look when you have a window? Happy to adjust anything further.

